PR #2, opened 2017-08-23
Use relative paths to user home instead of absolute

bre7:
Awesome idea Florent !

Just a minor fix to be able to open the workspace in any workstation if the user isn't `florent` 😃


flovilmart:
Hi, I'll be working on a side branch for the automatic parse-server startup, in the meantime, you can start playing with the playgrounds and start giving some feedback. Beware, that's a completely new philosophy, and I'm very open to changes :)

PR #6, opened 2017-09-24
Complete overhaul

flovilmart:
@pranjalsatija  what do you think?

- Removes RESTCommand, moves to API.Command
- All API calls are sync by default, to facilitate async and scheduling later
- Adds more options for queries like sessionToken, installationId, with the same template, we could also inject safely custom headers etc...
- API.Command is a value type now, no more subclassing for batch, just composition
- ParseErrors are safe to use (imported from the original SDK)
- All calls either succeed and return or throw, which makes it easier for tests etc...


pranjalsatija:
@flovilmart I think it looks great! Placing `Command` inside of `API` was a really good idea.

pranjalsatija:
Quick question. Will the SDK provide async interfaces for calls? Since everything is now synchronous by default, will it be up to the SDK's users to call things on the background thread or will we still build out async interfaces that wrap the synchronous methods?

flovilmart:
They are provided already :)

flovilmart:
Here: https://github.com/parse-community/Parse-Swift/blob/a4fe2ece8421c0693e7709bdf2095577c6120dbf/Sources/ParseSwift/Asynchronous.swift

pranjalsatija:
@flovilmart My bad, I didn't look hard enough. I'm in the middle of a bunch of stuff with exams and studying; sorry about that. Should I add `Installation` after you merge this?

flovilmart:
Focus on your exams! And yes you can probably get started on the Installation!

pranjalsatija:
Alright! I’ll get to work on it soon.

--
Thanks
Pranjal Satija
570.956.7064

PR #11, opened 2020-07-09
Upgrade swift and fix test cases

cbaker6:
Testing...


drdaz:
Heh.. I think we've probably done the same things :)

https://github.com/drdaz/Parse-Swift

PR #12, opened 2020-07-09
Upgraded to Swift 5

cbaker6:
This PR builds off of ShawnBaeks PR #9 as many of fixes seemed to be approved by Flow.

In addition to upgrading, I made changes to the force unwrap test cases as they should fail gracefully and allow the rest of the tests to run even if one fails.


TomWFox:
Great to see some movement here! I’ve converted the PR to a draft, feel free to mark as ready to review when you’ve finished working on it.

drdaz:
Yeah this is awesome sauce. I've spent a little time with this repo too. Could be fun to talk about where to go from here.

drdaz:
@cbaker6 fwiw I looked at the 'Any' KeychainStore test you seem to be working on. And while I found the problem interesting, I opted to leave it alone (after some time poking at it 😃).

As best I can tell the KeychainStore component is a direct port from the obj-c library. So we're taking the tests and assuming the requirements behind them are ideal. When I look at how the KeychainStore component is consumed in the obj-c library, it doesn't seem to actually make use of the Any case. And the method isn't exposed to clients.

So... I'm not certain that the functionality is required to build a Swift implementation of Parse around. That's not a suggestion not to do it, just sharing what I've seen.

cbaker6:
@drdaz From the original PR that @ShawnBaek posted it seems Flo had approved of AnyCodable. There was a similar discussion about Any and some form of AnyCodable on AlamoFire. My intuition is since this framework (the swift version anyway) is in it's beginning stages, it easier to add it in now than in Alomofire's case. 

It's almost working now, just a casting issue which I hope to have fixed soon. The reason the previous test didn't pass completely was an SPM issue

cbaker6:
@TomWFox this PR is ready. A few notes.. This PR brings the framework back to working order so others can start contributing again. A few limitations:

1. Currently this cannot be imported via SPM < 5.2. This has something to do with the file structure of "Sources" as it has trouble linking some of the files in 5.1 and below. IMO, this isn't a big deal as the older versions can still import directly or through Cocoapods and Carthage (at least via the tests anyway). Of course someone better at SPM than me might know how to fix this as it's probably an easy fix for someone who knows SPM. IMO, this isn't a showstopper for the PR

still writing...

ShawnBaek:
@cbaker6 Thanks a lot!!!

cbaker6:
@TomWFox np, I’m not sure what happened here, I accepted your changes, but it dismissed all of the reviews which wasn’t my intention

TomWFox:
Nothing to do with you, it's because of some branch protections I just added. I think I'll remove that particular one as it's a bit annoying!

PR #13, opened 2020-07-19
Renaming, Reorganization, and PrimitiveObjectStore

pranjalsatija:
This is a WIP, but I wanted to open it and start getting feedback. I still plan on:
* Moving protocol conformances in `ParseObject` around so that each protocol is implemented as an extension on `ParseObject.
* Doing the same in `ParseUser`.
* Nesting the types in `Coding/Miscellaneous.swift` under a `ParseCoding` namespace or something. Calls to the global `getJSONEncoder` and `getParseEncoder` functions feel weird.
* Adding a `PrimitiveObjectStore` protocol, extending `KeychainStore` to implement it, and using the new store to persist the current user.


cbaker6:
@pranjalsatija I skimmed this and it lgtm so far.

Just to let you know, we are working in parallel, I'm adding the unit testing in for Object, User, and login which all need URLMocking, so I'm adding that too. I'll try to finish it before yours so we can use it verify your changes keeps everything intact. My guess is we will touch very few if any of the same files, but will need to switch the unit tests I add over to ParseObject, etc.

pranjalsatija:
Simple storage is done, but not tested yet. I'll wait on @cbaker6's tests to land before I write any new ones so we don't step on each other's toes.

Also, while I'm doing cleanup / reorg stuff, a few more things than originally planned come to mind. @cbaker6 @TomWFox or anyone else, I'd like your thoughts:
* I don't love the `Equatable` [implementation](https://github.com/pranjalsatija/Parse-Swift/blob/ada568cb0a0d5800e01518b53e7f79ba65a40d19/Sources/ParseSwift/Objects/ParseObject.swift#L76) on `ParseObject`. I don't think that 2 objects with the same ID but different data should be considered unequal by the `==` operator. IMO if that's what users want, they can compare the IDs themselves pretty easily. Should we remove this?
* Should we remove [AnyCodable](https://github.com/Flight-School/AnyCodable)?
  * We don't use it in the library ourselves and it's quite a bit of code + tests.
  * I feel like this sort of thing isn't in our domain per se; if users need that kind of functionality, they can include AnyCodable directly as a Swift package and use it.
  * Having 0 dependencies in the core library would be cool.
  * The code was copied + pasted into this project, but AnyCodable is still actively maintained, and our version is definitely already behind the current one. There's a new release from 12h ago.
* I understand the rationale for [`ParseEncoder`](https://github.com/pranjalsatija/Parse-Swift/blob/rename/Sources/ParseSwift/Coding/ParseEncoder.swift), but I think there may be a better way to skip some coding keys than copying + pasting parts of `Foundation.JSONEncoder`. I can (probably) write a much simpler `ParseEncoder` from scratch that builds a simple Dictionary internally and then calls on `Foundation.JSONEncoder` to encode the dictionary.
  * This will almost certainly be less code. Our copy of `JSONEncoder` is more than 800 LOC.
  * We won't have to keep up with / update our copy of the encoder as the Foundation one changes over time.
  * I'm going to do this for fun and to see what it looks like no matter what, but I'd like to hear if anyone has a reason not to.
* I think we can do more with the protocols we have. What I mean by this is that despite the fact that we have separate protocols like `Saveable`, `Fetchable`, etc. the API itself is fairly brittle in the fact that it only accepts `ParseObject` instances.
  * For example, why should [`API.createCommand`](https://github.com/pranjalsatija/Parse-Swift/blob/3547627753e40429d43385b211ae6454109aecd8/Sources/ParseSwift/API/API%2BCommands.swift#L74) care if its input is a `ParseObject`? It should just require it to be `Saveable` and have an `updatedAt` field (maybe we can add an `Updatable` protocol or something to codify this). 
  * There's nothing explicitly *wrong* with the way things are now, but I think we can do better and write better, simpler, "Swiftier" code. It would also make unit testing things like `API` easier.

mtrezza:
> I don't love the Equatable implementation on ParseObject. I don't think that 2 objects with the same ID but different data should be considered unequal by the == operator. IMO if that's what users want, they can compare the IDs themselves pretty easily. Should we remove this?

I would say equality should be determined solely by object ID. That's the purpose of an "identifier". How are the Parse Android / JS SDK comparing, for consistency?

cbaker6:
@pranjalsatija I'm still working on the tests. I hope to have them done soon. Below is my input on some of the things you brought up:

> I don't love the Equatable implementation on ParseObject. I don't think that 2 objects with the same ID but different data should be considered unequal by the == operator. IMO if that's what users want, they can compare the IDs themselves pretty easily. Should we remove this?

I think it matters how parse and the other SDKs handle this. I agree with @mtrezza response and we can check the obj-c framework as well

> Should we remove AnyCodable?
I don't think we should be in a rush to remove this as it's not hurting anything right now. AnyCodable gives us the ability to do [Any] and since we are working with JSON this can definitely be valuable. If at some point in the future we find we don't need it, then I think we can remove it, but LOC for this I don't believe is an issue.

> Having 0 dependencies in the core library would be cool.
I agree here, but sometimes it may lead to more development time and a dependency may be optimized for the specific thing it's designed for. @mtrezza I believe what @pranjalsatija is referring to is a comment from Flo which said something along the lines with 0 dependencies. As we've seen with the obj-c framework, the Bolts dependency became a major hurdle. Flo pointed out that `Foundation` is powerful enough to handle everything Bolts was doing and that we should shoot for zero dependencies. I agree, `Foundation` is powerful and as of now, haven't seen where we specifically need a dependency, but that of course can change as this framework grows. I think we should shoot for zero, and if a possible dependency comes up, we should discuss.

> The code was copied + pasted into this project, but AnyCodable is still actively maintained, and our version is definitely already behind the current one. There's a new release from 12h ago.
I don't believe `AnyCodable` or `ParseEncoder` is intended to be maintained in this framework as a "dependency", more of, we cite the source of where it came from, and then we tailor the requirements to Parse and maintain those requirements. This will lead to @mtrezza point of reducing the LOC if we don't need things which I'm sure there may be parts of AnyCodable we don't need. When making updates to AnyCodable from when it was added a few years ago, I didn't pull in all of the changes they made. I also made some updates because they force unwrap in some places I don't agree with. As with `ParseEncoder`, we should "Parsify" AnyCodable and not treat as a dependency that will be updated by external maintainers
 

> I understand the rationale for ParseEncoder, but I think there may be a better way to skip some coding keys than copying + pasting parts of Foundation.JSONEncoder. I can (probably) write a much simpler ParseEncoder from scratch that builds a simple Dictionary internally and then calls on Foundation.JSONEncoder to encode the dictionary.

I haven't looked into ParseEncoder in a way of thinking of optimizing it, but merely to get everything up to spec. I also "assumed" that it worked from the manual tests. I do know Flo cited the JSON encoder for the ParseEncoder design. I don't think he intended to keep it fully aligned with the JSONEncoder, but merely "Parsify" what was there. I do think it's reasonable to improve code, but I think we should establish some barriers before we attempt to make major changes. For example, there's no test cases around this. IMO if we want to make a major change in an early stage of framework, we should write the unit tests for the current working cases (maybe some of the unit tests show the limitations of the current) and then write the new version and run those same tests against it. By doing major overhauls without unit tests we can run into big problems with breaking code

pranjalsatija:
@mtrezza 
> I would say equality should be determined solely by object ID. That's the purpose of an "identifier". How are the Parse Android / JS SDK comparing, for consistency?

I looked at [`PFObject.h`](https://github.com/parse-community/Parse-SDK-iOS-OSX/blob/master/Parse/Parse/PFObject.h) in the ObjC SDK and there's no implementation of `isEqual`. By default, `NSObject` just checks pointer equality. In addition, semantically, Swift value types should not have identity and just be dumb containers for data. I think it breaks convention if you have 2 objects with different data in them but they compare unequal. Comparing IDs directly is more clear, and if we override equality this way, we have no way left to see if objects are fully equal. Swift also has `Identifiable` now for IDing purposes, so maybe we can implement that for object identification?

> Makes sense. I think an important factor here is performance, whatever the solution is, especially when it comes to large objects / arrays with thousands of entries. In Parse Server, JSON encoding in the storage adapter is a significant bottleneck.

I agree. I'll do what @cbaker6 said and write some unit tests for the current `ParseEncoder`, with performance benchmarks built in. Then, once the new one is done, we can see how it does.

> Could we use it to reduce our LOC? I don't think a 0 dependency SDK is a goal to strive for per se if a dependency can be expected to be well maintained in the future and it reduces our code.

@cbaker6 mentioned this, but the original idea was that relying on Bolts and other stuff ended up becoming a problem later on. I don't personally see where / how we could use `AnyCodable` in the framework at the moment, but I'm not opposed to revisiting this when we do the first release.

cbaker6:
@pranjalsatija a quick note about `Identifiable`, you are correct in it's usage, the problem us is that it's iOS13 and up so it will limit us. I think comparing the objectId is okay for us

pranjalsatija:
@cbaker6 There's a lot of overlap in what you and @mtrezza said, so I'll just address a few of the things you mentioned:

> I don't think we should be in a rush to remove this as it's not hurting anything right now...

That's fine with me. We can leave it in until we have a release, then evaluate whether or not we're using it and decide later.

> I don't believe AnyCodable or ParseEncoder is intended to be maintained in this framework as a "dependency"...

This makes a lot of sense. We can adapt them however we need to, and again, if we find we don't need them after building out some functionality, we can get rid of them then. As for `ParseEncoder`, I think that writing an encoder that's capable of spitting out a Dictionary would be useful. We can maybe use `AnyCodable` here? This would give us the ability to do things like only send "dirty" keys to the server, the way the ObjC framework does. We don't have the same dynamic capabilities in Swift, but having an encoder that can give us a Dictionary would probably be a good substitute.

>  I haven't looked into ParseEncoder in a way of thinking of optimizing it, but merely to get everything up to spec. I also "assumed" that it worked from the manual tests...

I'll write some tests for it. Should I wait until after your tests are up? Regardless, I'll test it and include benchmarks so we can compare the 2 implementations. In the mean time, I was able to cut down on some duplication by reusing some stuff from `JSONEncoder` instead of having a separate copy in our repo:

```swift
public typealias DataEncodingStrategy = JSONEncoder.DataEncodingStrategy
public typealias DateEncodingStrategy = JSONEncoder.DateEncodingStrategy
public typealias OutputFormatting = JSONEncoder.OutputFormatting
public typealias NonConformingFloatEncodingStrategy = JSONEncoder.NonConformingFloatEncodingStrategy
```

mtrezza:
@pranjalsatija 

> Comparing IDs directly is more clear, and if we override equality this way, we have no way left to see if objects are fully equal.

I think our positions are `equal` ;) I would also go for solely comparing by ID to determine equality, if I understand your point correctly. If two objects have the same ID but different data (because one of them has been modified), then a "dirty" property should indicate that one object has been modified. I would see it as the dev's job to differentiate between "equal" and "dirty".

pranjalsatija:
@mtrezza Sweet! Just to clarify with you and @cbaker6, we're OK with removing the `==` override, right? And letting users compare IDs directly?

Just to recap, this is what I plan on doing next and including in this PR (we really need a board or some other way to track / assign tasks):
* (If we all agree) remove the custom `Equatable` implementation on `ParseObject`. I'll add an `isSameObject` or something similar as a convenience for comparing IDs, or users can do that directly. In addition, I'll add `Identifiable` conformance on platforms where it's available.
* Write good unit tests for `ParseEncoder`, including some performance benchmarks.
* Write a new `ParseEncoder` that encodes to a `Dictionary` internally and then leverages `Foundation.JSONEncoder` to produce JSON. We'll also leave the option to "encode" to a Dictionary open for usage, and maybe we can use that functionality later on. If performance is acceptable, we can replace the current `ParseEncoder` with the new one.

mtrezza:
> we're OK with removing the == override, right? 

Not sure about that, as I said before, how does the Parse Android / JS SDK handle this, so we can be consistent?

pranjalsatija:
JS has an `equals` method that only compares object ID and class name.

Android and ObjC do more or less what I'm proposing:
* Android has a `hasSameId` method that compares class name and object ID, but no implementation for the standard Java `equals`.
* ObjC's `isEqual` and `==` both do pointer equality, and it's on the user / framework to do other comparisons if they want.

mtrezza:
> I think having a hasSameID method or something is most consistent with the other mobile frameworks, and also most consistent with Swift conventions.

I like it.

Sources/ParseSwift/Coding/Extensions.swift
```diff
@@ -0,0 +1,37 @@
+//
+//  File.swift
+//
```
cbaker6, line 3: Switch file name in comments to match file

Sources/ParseSwift/Coding/ParseCoding.swift
```diff
@@ -0,0 +1,80 @@
+//
+//  ParseObjectType.swift
```
cbaker6, line 2: Switch comment file name to match new file name

Sources/ParseSwift/Parse Types/BaseParseUser.swift
```diff
@@ -0,0 +1,20 @@
+//
+//  File 2.swift
```
cbaker6, line 2: File name update

Sources/ParseSwift/Parse Types/BaseParseUser.swift
```diff
@@ -0,0 +1,20 @@
...
+//  
+//
+//  Created by Pranjal Satija on 7/19/20.
+//
+
+import Foundation
+
+/// Used internally to form a concrete type representing `ParseUser`.
+/// `ParseUser` itself has associatedtype requirements, so it's often awkard to use it directly.
+struct BaseParseUser: ParseUser {
```
cbaker6, line 12: Can you explain why this would be needed? Ir looked like the UserType that had the CurrentUserContainer and LoginInfo was handling this fine

Sources/ParseSwift/Storage/PrimitiveObjectStore.swift
```diff
@@ -0,0 +1,53 @@
+//
+//  File.swift
```
cbaker6, line 2: File name update

cbaker6:
> For example, why should API.createCommand care if its input is a ParseObject? It should just require it to be Saveable and have an updatedAt field (maybe we can add an Updatable protocol or something to codify this).

I’m thinking there’s reasoning behind this. It seems that only ParseObjects or something derived from a ParseObject should be pushed to the parse-server for saving. I didn’t see anything in savable that ensures this, so then you will allow pushing anything that conforms to savable that isn’t a ParseObject which the parse-server will not know how to save because it may not contain the correct attributes

cbaker6:
> I understand the rationale for ParseEncoder, but I think there may be a better way to skip some coding keys than copying + pasting parts of Foundation.JSONEncoder. I can (probably) write a much simpler ParseEncoder from scratch that builds a simple Dictionary internally and then calls on Foundation.JSONEncoder to encode the dictionary.

I would like to see how this comes out when you design it. It would definitely benefit from using apples maintained JSONEncoder.

Sources/ParseSwift/Coding/Extensions.swift
pranjalsatija: Good catch! I'll fix all of these up.

Sources/ParseSwift/Parse Types/BaseParseUser.swift
pranjalsatija: On second thought, it's not needed. I can use `AnyCodable` instead of doing tricky stuff with generics and this type. I'll make that change tomorrow!

pranjalsatija:
> I’m thinking there’s reasoning behind this. It seems that only ParseObjects or something derived from a ParseObject should be pushed to the parse-server for saving...

I agree, but then maybe the `Saveable` protocol itself is redundant? What I actually had in mind was expanding each of the protocols. For example, `Saveable` would expand to the following:

```swift
protocol Saveable {
  var endpoint: API.Endpoint { get }
  var isSaved: Bool { get set }

  func save(options: API.Options) throws -> SavingType
  func save() throws -> SavingType
}
```

And then we could provide default implementations for `save(...)`, using `endpoint`, `isSaved`, etc. to automatically create the necessary API commands. I like this approach because I feel like it takes full advantage of the fact that the protocol exists, which our current implementation doesn't. I also like the idea of encapsulating saving functionality this way, for what it's worth; code related to saving things would only care about the small slice of information necessary to save them, and this would almost definitely make mocking for unit tests easier.

On the other hand, we'll only ever have a few types that implement `Saveable` and friends. If that's the case, do we need the protocols at all? Will there ever be an API that returns an `[Saveable]`, for example? The way things currently are, we don't actually do anything with any of the protocols. They exist but mostly serve as eye candy; there's no code that accepts an `[Saveable]` or something like that. If we continue down this route, we could just get rid of them.

I'm open to either approach. I just think that if the protocols exist in the codebase, we should leverage them in some useful way.

> Essentially you can use the test cases I added as I use the ParseEncoder heavily, but you would need to make a few changes that I know of. When using the ParseEncoder, there’s still something fishy going on with the dates.

Sounds good, thank you for the heads up! I'll be extra watchful about date coding when I get around to implementing this (hopefully tomorrow, but not 100% sure).

cbaker6:
> Sounds good, thank you for the heads up! I'll be extra watchful about date coding when I get around to implementing this (hopefully tomorrow, but not 100% sure).

I take back my comments about the Date encoder not working. I wouldn't pass when I was previously testing. I moved some stuff around, it worked, I moved it back, it still works (weird). Anyways, I made the tests I added in PR 14 compare the encoded/decoded dates directly, and it's passing in travis. So the changes you make to the encoder will also be tested by the other unit tests

cbaker6:
> I take back my comments about the Date encoder not working. It wouldn't pass when I was previously testing. I moved some stuff around, it worked, I moved it back, it still works (weird). Anyways, I made the tests I added in PR 14 compare the encoded/decoded dates directly, and it's passing in travis. So the changes you make to the encoder will also be tested by the other unit tests

I found out a little more about this and why I was having the issue with encoding/decoding dates. The issues are only with `.fetch`. Since that was the first test case I implemented, I thought it was the case for update, save, etc. Skimming through `.fetch` it doesn't take the exact same steps as the others when decoding and doesn't have a `.apply`. I don't know the specifics of this, but this should give you a step in the direction for a fix. If you look at the test cases, these are the only ones I check for NotNil instead of == the dates because the strings look similar but something minor seems to be different

Sources/ParseSwift/Objects/ParseObject.swift
```diff
@@ -0,0 +1,134 @@
+//
+//  ParseObject.swift
+//  ParseSwift
+//
+//  Created by Pranjal Satija on 7/18/20.
```
cbaker6, line 5: It looks like many of these files were renamed, but the Flo’s name was removed and the original date it was created updated to the current date. Are you creating new files and dropping in the code or renaming them in Xcode? If you are creating new files, I think you are going to have merge conflicts with the other PRs

I think the original creator and dates should be left and if there needs to be additional comments, we can add them. If you create new files with new code, then it should look the way this file looks

An example I had is in PR 15 was URLSession+sync.swift. I had created URLSession+async.swift (brand new file and code), after the bugs were worked out, I deleted URLSession+sync.swift as it wasn’t being used anymore and renamed URLSession+async.swift to URLSession.swift. This case is slightly different, but I’m even thinking of putting Flo’s info back in there.

TomWFox: Just to add here... the copyright to Parse should be changed to Parse Community or removed at some point.

And in terms of who's name is in there, GitHub holds all the history so I really don't think it's important / helpful to have names at all?

cbaker6: I’m not by my computer to verify, but I don’t believe the file will show the history if it’s created as new with a different name. I think it will still shows if you rename the file, or if a file is deleted and then a new file is created in the future with the same name. For example I don’t think the URLSession.swift file will show the history based on what I mentioned in the previous post. My belief is we should always cite the source and preserve the history when possible.

@TomWFox to your point about the names, I think the Parse-Swift Xcode project is currently setup to automatically input our names, copyright, and date. For example, I haven’t intentionally add my names to any files in the PRs, it was done automatically.

pranjalsatija:
Sorry for the delay on this! I'm currently on vacation but looking for time to get some stuff done.

Sources/ParseSwift/Objects/ParseObject.swift
pranjalsatija: @cbaker6 Good catch, I'll put Flo's name back. I don't use the file rename built into Xcode because I've found it to be unreliable sometimes. I usually create a new file, and then copy/paste the body from import statements down. Xcode just automatically inserts my name and copyright at the top. It's never something I really noticed or paid a lot of attention to, but I'll fix it.

@TomWFox Xcode inserts those names and copyrights by default. If it's an issue, I think we can follow the steps listed in [this](https://useyourloaf.com/blog/changing-xcode-header-comment/) article to insert a custom copyright template for this project that includes an abridged license or any other information, along with a copyright owned by Parse Community.

pranjalsatija:
Just a heads up, I'm back and working on this now. Sorry again for the delay.

Sources/ParseSwift/Objects/ParseObject.swift
pranjalsatija: @cbaker6 @TomWFox what do you guys think about removing these headers entirely? Git already keeps track of who made what, when they made it, etc. and these headers will quickly get cumbersome to maintain. If nobody has a problem with it, I can configure Xcode not to make them for new files, and I can get rid of the ones that are already there.

Sources/ParseSwift/Parse Types/BaseParseUser.swift
pranjalsatija: @cbaker6 so I've been trying to find a good solution to this, and I honestly think the `BaseParseUser` solution is the best one. Here's why:

* `AnyCodable` doesn't work unless the type you're encoding is one of a few predefined ones (see the implementation for `encode(to: Encoder)` in `AnyEncodable.swift`. I didn't know this when I said `AnyCodable` would solve the issue here.
* Without `AnyCodable`, `CurrentUserContainer` has to be generic because `ParseUser` has associated type and Self requirements, so we can't type something as `ParseUser` directly.
* If `CurrentUserContainer` is generic, then we need a concrete type within the SDK that we can use to instantiate a `CurrentUserContainer` and get the auth token. Outside of the SDK, it's not necessary, because SDK users will have their own concrete implementations of `ParseUser`.

All of these things put together make it pretty hard to store an arbitrary `ParseUser`. I think the `BaseParseUser` solution is good because when we're *setting* the current user, we can use the concrete type of `ParseUser` that's available when signing up, logging in, etc. When we're *getting* it, we can use `BaseParseUser` because anything that conforms to `ParseUser` should be decodable to a `BaseParseUser`.

Please let me know if you have any other ideas!

cbaker6:
Heads up, I forgot to pipefails in Google actions for the builds using Xcode tools. It looks like Carthage was catching the errors on your end, but you can see the specific ones in `build-macos` under "Build" (ignore the green checkmark). My guess is that you forgot to add your new files to all of the targets; macOS, watchOS, etc.

pranjalsatija:
@cbaker6 thank you for the heads up! I was planning on looking into that when everything was 100% done.

pranjalsatija:
All the current tests are passing.

Sources/ParseSwift/Objects/ParseObject.swift
pranjalsatija: @cbaker6 @TomWFox just want to mention this again before this PR is merged; if we want, we can save this change for another one down the road.

pranjalsatija:
@cbaker6 this is ready for a review!

pranjalsatija:
Fixed the tests. They broke because of a *really* weird issue. Basically, I think the compiler was optimizing some `Int` values into `Builtin.Int1` values because those ints had the value `1`. This made it possible to decode them into `Bool` instances (or maybe Codable always allows decoding `1` to `true`?) and JSON values with the number `1` were being incorrectly treated as bools, which caused obvious decoding errors.

I fixed this by updating `AnyEncodable` and `AnyDecodable` to always try encoding/decoding more complex types first, and then falling back on simpler ones like `Bool`. All the tests pass again, with no changes other than renaming.

pranjalsatija:
In addition, can we merge this before starting new PRs? It's really cumbersome to merge PRs like #14 and #15 into this branch, update them to account for renamed types, fix all the tests, etc.

cbaker6:
I’m going to try to look this over tonight

cbaker6:
@pranjalsatija I didn’t look at detail to the code you, but your last fix makes sense to me.

When it comes to the new encoder, what pro’s/con’s do you see vs the original ParseEncoder? This will help me when assessing the changes. Are there any limitations with either approach?

pranjalsatija:
I don't think there are a lot of cons. For what it's worth, all the tests passed perfectly fine with no changes, so the new encoder works as a drop-in replacement for everything we do with it in this framework. As far as pros go, we have:
* Less code. The new encoder is under 300 LOC, the last one was 800+.
* The ability to encode to a Dictionary. This will be useful later on when we implement a local store backed by SQLite. Since Swift objects aren't dynamic, we can't get all their properties at runtime like the ObjC framework can. The new encoder makes that possible so we can build tables, indexes, etc. in SQLite, even when users' model types are Swift structs.
* Dependence on the Foundation JSON encoder instead of a copy that we keep in our source code.

Cons:
* Maybe performance? I forgot to write benchmarks, but I can do that so we can see if there's a performance hit.

pranjalsatija:
So I just wrote a quick benchmark (see the `perftests`) branch on my repo, and here's what I saw. Note that each "invocation" is encoding 5 instances of `GameScore`:

* The new encoder is 2.5x slower for the very first invocation (0.001492 seconds vs. 0.000603 seconds). Not sure why this is, but I'll chalk it up to warmup time / getting CPU branch prediction working / something else.
* For subsequent invocations, it's between 20% and 60% slower. This number varies wildly from run to run, but we're looking at numbers like 0.000106 seconds per invocation vs. 0.000091 seconds per invocation.

Long story short, both encoders are really fast, taking something like 0.02 milliseconds to encode an object, but the new one is a bit slower.

cbaker6:
Both of those numbers sound good to me. With the addition of gaining all of the updates from the native encoder Along with any subsequent updates, I will say yours is a win if it can do everything the old one does. It’s possible the native encoder is slower than the old native encoder the original ParseEncoder was derived from, maybe because it does more? (Just a guess).

pranjalsatija:
I think the main issue is the double encoding. The new encoder first encodes to a Dictionary internally, then wraps the Dictionary in an AnyCodable, and then calls on the Foundation JSON encoder to encode the AnyCodable instance. I'd say most of the slowdown is probably in the first step.

Sources/ParseSwift/Coding/ParseEncoder.swift
```diff
@@ -0,0 +1,259 @@
...
+            skippingKeys: skippedKeys
+        )
+    }
+
+    static func encode<T: Encodable>(
+        _ value: T,
+        with codingPath: [CodingKey],
+        skippingKeys skippedKeys: Set<String>
+    ) throws -> Any {
+        switch value {
```
cbaker6, line 93: Looks like my NSMutable speculation might be true?

Sources/ParseSwift/Coding/ParseEncoder.swift
```diff
@@ -0,0 +1,259 @@
...
+    mutating func nestedUnkeyedContainer(forKey key: Key) -> UnkeyedEncodingContainer {
+        _ParseEncoderUnkeyedEncodingContainer(
+            codingPath: codingPath + [key],
+            dictionary: dictionary,
+            skippingKeys: skippedKeys
+        )
+    }
+
+    mutating func superEncoder() -> Encoder {
+        fatalError()
```
cbaker6, line 157: Maybe fill in the fatal error reasoning here? "Error, not allowed to call superEncoder". 

Are you able to mark these as `@available(*, unavailable)`?

Sources/ParseSwift/Coding/ParseEncoder.swift
```diff
@@ -0,0 +1,259 @@
...
+
+    mutating func superEncoder(forKey key: Key) -> Encoder {
+        fatalError()
+    }
+}
+
+// MARK: _ParseEncoderSingleValueEncodingContainer
+internal struct _ParseEncoderSingleValueEncodingContainer: SingleValueEncodingContainer {
+    let codingPath: [CodingKey]
+    let dictionary: NSMutableDictionary
```
cbaker6, line 168: Same question here with NS

Sources/ParseSwift/Coding/ParseEncoder.swift
```diff
@@ -0,0 +1,259 @@
...
+    }
+}
+
+// MARK: _ParseEncoderUnkeyedEncodingContainer
+internal struct _ParseEncoderUnkeyedEncodingContainer: UnkeyedEncodingContainer {
+    let codingPath: [CodingKey]
+    let dictionary: NSMutableDictionary
+    let skippedKeys: Set<String>
+
+    var array: NSMutableArray {
```
cbaker6, line 196: same here with NS, also the forced casting doesn't seem like the best approach. If we can't cast, we should be throwing, errorring with feedback and not crashing, or worse case fatalError

Sources/ParseSwift/Coding/ParseEncoder.swift
```diff
@@ -0,0 +1,259 @@
...
+    init(codingPath: [CodingKey], dictionary: NSMutableDictionary, skippingKeys: Set<String>) {
+        self.codingPath = codingPath
+        self.dictionary = dictionary
+        self.skippedKeys = skippingKeys
+
+        self.array = NSMutableArray()
+    }
+
+    mutating func encodeNil() throws {
+        array.add(NSNull())
```
cbaker6, line 222: NSNull? I haven't really dug into the Foundation encoder, but is this typical? I would have thought they would have moved away from these types of things. I can see it being used in Unarchiver/Archiver

Sources/ParseSwift/Parse Types/BaseParseUser.swift
cbaker6: Seems reasonable, I don't know how I feel about the `Base` part, but I currently can't think of something better.

I'm going to look back at my previous comment to remove it as I was looking for this exchange at the beginning of the review

Sources/ParseSwift/Storage/ParseStorage.swift
```diff
@@ -0,0 +1,47 @@
+//
+//  File.swift
```
cbaker6, line 2: Nit on the filename

Sources/ParseSwift/Storage/PrimitiveObjectStore.swift
cbaker6, line 2: Nit on the filename

Tests/ParseSwiftTests/ParseObjectBatchTests.swift
```diff
@@ -136,7 +136,7 @@ class ParseObjectBatchTests: XCTestCase { // swiftlint:disable:this type_body_le
             switch saved[0] {
 
             case .success(let first):
-                XCTAssertEqual(first, scoreOnServer)
+                XCTAssert(first.hasSameObjectId(as: scoreOnServer))
```
cbaker6, line 139: I'm not sure what happened here? The ParseObjects had something that already compared the objects directly by using ==. I know there was a discussion with @mtrezza about how the other SDKs handled it. It seems like we shouldn't require the developers to call `hasSameObjectId` and should do it for them as one may naturally attempt to compare two objects using ==

Sources/ParseSwift/Coding/ParseEncoder.swift
```diff
@@ -0,0 +1,259 @@
...
+//  Copyright © 2020 Parse. All rights reserved.
+//
+
+// This rule doesn't allow types with underscores in their names.
+// swiftlint:disable type_name
+
+import Foundation
+
+// MARK: ParseEncoder
+public struct ParseEncoder {
```
cbaker6, line 15: I think this should be marked `internal`, even when an external framework conforms to the storage protocol they should ask ParseSwift to encode/decode. This gives us the flexibility of modifying the ParseEncoder as we see fit as leaving it public enables developers to think they can override and make their own versions of the ParseEncoder

cbaker6:
@pranjalsatija overall I thought your changes looked good. Had a few nits and a few questions/suggestions

Sources/ParseSwift/Storage/ParseStorage.swift
```diff
@@ -0,0 +1,47 @@
...
+    public static var shared = ParseStorage()
+
+    private var backingStore: PrimitiveObjectStore!
+
+    mutating func use(_ store: PrimitiveObjectStore) {
+        self.backingStore = store
+    }
+
+    private mutating func requireBackingStore() {
+        guard backingStore != nil else {
```
cbaker6, line 19: I don't think we should create the inMemory store for them, if they don't supply one, we should fatalError and tell them to supply one. We can give them the option to use the one you created, but setting it up for them enables developers to not pay attention and then complain about the storage using memory

cbaker6:
In the spirit of code coverage, it looks like you can bump the `target` in .codecov.yml to 50. If you do add the tests I mentioned (if they are not there already), you maybe able to bump it higher to encourage future contributors not to add something without tests. I've been putting this as -2% of the current code coverage whenever I make an addition.

ParseSwift.playground/Pages/2- Finding Objects.xcplaygroundpage/Contents.swift
```diff
@@ -10,7 +10,7 @@ PlaygroundPage.current.needsIndefiniteExecution = true
 
 initializeParse()
 
-struct GameScore: ParseSwift.ObjectType {
+struct GameScore: ParseSwift.ParseObject {
```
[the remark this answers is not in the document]
pranjalsatija: I don't have a problem with `ParseSwift.Object`, but doing so would require making `Object` a nested typealias of `ParseSwift`, and that would make both `Object` and `ParseSwift.Object` valid.

I think it would be better to just have `ParseObject`, but I'm not against the other approach.

ParseSwift.playground/Pages/3- Users.xcplaygroundpage/Contents.swift
```diff
@@ -7,7 +7,7 @@ PlaygroundPage.current.needsIndefiniteExecution = true
 import ParseSwift
 initializeParse()
 
-struct User: ParseSwift.UserType {
+struct User: ParseSwift.ParseUser {
```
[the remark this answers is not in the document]
pranjalsatija: See above for my main comments on this, but `Object` and `User` will both be available outside the framework, no? When users want to define their own objects, they'll need to mark them as conforming to `ParseObject` and `ParseUser` (or whatever we end up calling them).

Sources/ParseSwift/Coding/Extensions.swift
```diff
@@ -0,0 +1,41 @@
...
+//  File.swift
```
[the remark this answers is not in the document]
pranjalsatija: Good catch. These headers will be the death of me; I'll fix all of these.

Sources/ParseSwift/Coding/ParseCoding.swift
```diff
@@ -0,0 +1,75 @@
...
+    static let jsonDateEncodingStrategy: JSONEncoder.DateEncodingStrategy = .custom(parseDateEncodingStrategy)
+
+    static let parseDateEncodingStrategy: AnyCodable.DateEncodingStrategy = { (date, enc) in
+        var container = enc.container(keyedBy: DateEncodingKeys.self)
+        try container.encode("Date", forKey: .type)
+        let dateString = dateFormatter.string(from: date)
+        try container.encode(dateString, forKey: .iso)
+    }
+
+    static let dateDecodingStrategy: JSONDecoder.DateDecodingStrategy = .custom({ (dec) -> Date in
```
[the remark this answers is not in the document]
pranjalsatija: They don't; these were originally written by Flo and just got marked as edited when I shuffled them around. I'll fix.

Sources/ParseSwift/Coding/ParseCoding.swift
```diff
@@ -0,0 +1,75 @@
...
+        do {
+            let container = try dec.singleValueContainer()
+            let decodedString = try container.decode(String.self)
+            return dateFormatter.date(from: decodedString)!
```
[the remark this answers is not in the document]
pranjalsatija: This was also written by Flo. I'll dig into what's going on here, because something about the error handling feels wonky.

Sources/ParseSwift/Coding/ParseEncoder.swift
```diff
@@ -0,0 +1,259 @@
...
+        let encoder = _ParseEncoder(codingPath: [], dictionary: NSMutableDictionary(), skippingKeys: skippedKeys)
+        try value.encode(to: encoder)
+
+        // swiftlint:disable:next force_cast
+        return encoder.dictionary as! [AnyHashable: Any]
+    }
+
+    func encode<T: Encodable>(_ value: T) throws -> Data {
+        let dictionary = try encodeToDictionary(value)
+        return try jsonEncoder.encode(AnyCodable(dictionary, dateEncodingStrategy: dateEncodingStrategy!))
```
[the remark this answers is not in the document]
pranjalsatija: Good catch. There's no need to unwrap at all here, I updated `AnyCodable` to take an optional param but forgot to replace it here. I'll fix.

Sources/ParseSwift/Coding/ParseEncoder.swift
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// MARK: _ParseEncoder
+internal struct _ParseEncoder: Encoder {
+    let codingPath: [CodingKey]
+    let dictionary: NSMutableDictionary
+    let skippedKeys: Set<String>
+    let userInfo: [CodingUserInfoKey: Any] = [:]
+
+    init(codingPath: [CodingKey], dictionary: NSMutableDictionary, skippingKeys: Set<String>) {
```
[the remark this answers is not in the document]
pranjalsatija: So `NSMutableDictionary` and `Dictionary` actually have very different behavior; the `NS` version is a reference type, and it's still around for cases where you actually want reference types. In this case, we need it to work as a reference type because the dictionary gets passed around to a few different places as encoding occurs, and we want everyone writing into the same dictionary.

Sources/ParseSwift/Coding/ParseEncoder.swift
pranjalsatija: Can you clarify? I provided a comment about NS vs. the native Swift types above, but I'm not sure how it applies to this chunk of code.

Sources/ParseSwift/Coding/ParseEncoder.swift
pranjalsatija: I can't do the `unavailable` thing because then the compiler complains about an unavailable method being used to satisfy protocol requirements. I'll add error messages though.

Sources/ParseSwift/Coding/ParseEncoder.swift
pranjalsatija: We can't `throw` in a computed property, but I'll add a `fatalError` with a message.

I'd say if this code path ever executes, it's a serious problem because that means someone else tried to encode the same thing into the same dictionary with a different type, *after* this unkeyed coding container was called to do the same thing, and that's definitely a crash-worthy error.

Sources/ParseSwift/Coding/ParseEncoder.swift
pranjalsatija: You can't add `nil` to an `NSArray`, so you have to add `NSNull` instead.

Tests/ParseSwiftTests/ParseObjectBatchTests.swift
pranjalsatija: Maybe there's a misunderstanding about that discussion. My thought was that we should leave equality and this equality-esque thing separate, and my goal was to leave `==` for cases where we want to check strict, deep, object equality, if that ever comes.

Sources/ParseSwift/Coding/ParseEncoder.swift
pranjalsatija: Sounds good, I'll make it `internal`.

Sources/ParseSwift/Storage/ParseStorage.swift
pranjalsatija: I'll update this. We can move it to the `initialize` call that users have to make when providing an app ID and stuff.

pranjalsatija:
The progress on tests is awesome! As for testing `PrimitiveObjectStore`, it's a protocol, so I can't test it directly. I extended `KeychainStore` to conform to it, and I believe `KeychainStore` is already tested.

ParseSwift.playground/Pages/2- Finding Objects.xcplaygroundpage/Contents.swift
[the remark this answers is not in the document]
cbaker6: I’m on my iPad, so I can’t confirm this, but it sounds like how currently have it setup, `ParseObject` is valid now? If so, then I take back my comments as we can just encourage developers to use `ParseObject`and `ParseUser`. I think from our previous discussion we were trying to stay away from `Object` and `User`  as they are too generic and can easily conflict with another framework when made public

Tests/ParseSwiftTests/ParseObjectBatchTests.swift
cbaker6: Are you saying that == is still there? I assumed it was removed because these test cases were changed  from == to use `hasSameObjectId`. Did you only do this on a few to see if hasSameObjectId was working and left the rest of the == comparisons?

ParseSwift.playground/Pages/2- Finding Objects.xcplaygroundpage/Contents.swift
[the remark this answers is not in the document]
cbaker6: Nvm, I see you demonstrated it in your fixes. I think we are good with ParseObject and ParseUser

Sources/ParseSwift/Coding/ParseEncoder.swift
```diff
@@ -154,11 +154,11 @@ internal struct _ParseEncoderKeyedEncodingContainer<Key: CodingKey>: KeyedEncodi
     }
 
     mutating func superEncoder() -> Encoder {
```
cbaker6, line 156: Are you not able yo mark these as unavailable?

Sources/ParseSwift/Coding/ParseEncoder.swift
cbaker6: My comments here from earlier in reference to value for NSMutable, but I’m fine with your earlier response about the need for a reference and not making copies.

Tests/ParseSwiftTests/ParseObjectBatchTests.swift
pranjalsatija: No, there's currently no implementation for `==`. I replaced all the test cases that were doing equality with `==` with `hasSameObjectId`, because the implementation for `hasSameObjectId` is doing what the old `==` implementation did. At some point in the future, we'll have to provide a new implementation for `==` that does a real equality check.

cbaker6:
What are your thoughts on the ParseEncoder tests? From a coverage standpoint it’s at 53%. If you look at the file in Xcode it should show what parts aren’t being covered

https://codecov.io/gh/parse-community/Parse-Swift/compare/8d10c1cca91131b8c1055227036cabc0852419ed...8189db8c1cdc41a64631cc76943b7b4ca586c188/tree/Sources/ParseSwift/Coding

Sources/ParseSwift/Coding/ParseEncoder.swift
pranjalsatija: Nope. The compiler produces the following error:

```
Unavailable instance method 'superEncoder()' was used to satisfy a requirement of protocol 'UnkeyedEncodingContainer'
```

Tests/ParseSwiftTests/ParseObjectBatchTests.swift
cbaker6: Hmm, I’m not sure I agree with this. The objects with the same objectId should be == imo. Now if one or many of those object are dirty, they are still the same object, but “may” have different/outdated data. At the end of the day, that should get resolved at the parse-server level and depending on the use-case, the developer will add Cloud Code to resolve. When the resolution/synchronization occurs at the server level, all objects will be hydrated back with the new/same data (assuming there’s an Internet connection of course). So objects with the same objectId are the same, rather they are in sync is a different question. This goes into the data store convo we were originally having with @drdaz

pranjalsatija:
Here's what the report is showing as untested:
### ParseEncoder
* `ParseEncoder.singleValueContainer` - makes sense. This would only be used if we tried to use `ParseEncoder` to encode a simple scalar like `5`. I don't ever see us doing something like `ParseEncoder.encode(5)`, but I can add a test. It would get encoded as `{ "<root>": 5 }`.

### _ParseEncoderKeyedEncodingContainer
* `_ParseEncoderKeyedEncodingContainer.encodeNil` - This never gets called because the generic `encode<T>` gets called. The protocol definition for `KeyedEncodingContainerProtocol` actually lists out all the encodable types as separate methods, but a single generic type can cover them all. For some reason, the compiler still requires an explicit definition for `encodeNil`. I don't see how we could write tests that cover this at all, because `encode<T>` will always get called.
* `_ParseEncoderKeyedEncodingContainer.nestedContainer` - This could definitely use a test; I will add it.
* `_ParseEncoderKeyedEncodingContainer.nestedUnkeyedContainer` - Same here.
* `_ParseEncoderKeyedEncodingContainer.superEncoder` - We don't support it.
* `_ParseEncoderKeyedEncodingContainer.superEncoder(forKey:)` - We don't support it.

### _ParseEncoderSingleValueEncodingContainer
* All of `_ParseEncoderSingleValueEncodingContainer` - We rely on `AnyCodable` to encode single values within an object, so this should never get used for the same reason that `ParseEncoder.singleValueContainer` should never get used. However, it'll be covered by the test I add.

### _ParseEncoderUnkeyedEncodingContainer
* The `fatalError` in `_ParseEncoderUnkeyedEncodingContainer.array` - This should never happen anyway, so that seems right.
* `_ParseEncoderUnkeyedEncodingContainer.count` - We don't use this at all, it's just a protocol requirement I had to add.
* `_ParseEncoderUnkeyedEncodingContainer.encodeNil` - See the above comment about `_ParseEncoderKeyedEncodingContainer.encodeNil`.
* `_ParseEncoderUnkeyedEncodingContainer.nestedContainer` - This would get used when arrays of complex objects are in play. I'll add a test for it.
* `_ParseEncoderUnkeyedEncodingContainer.nestedUnkeyedContainer` - This would get used when nested arrays are in play. I'll add a test for it.
* `_ParseEncoderUnkeyedEncodingContainer.nestedUnkeyedContainer.superEncoder` - See above.

I'll add tests for the parts that make sense. It should get us well beyond 53% but not quite 100%.

Tests/ParseSwiftTests/ParseObjectBatchTests.swift
cbaker6: The sync comments are above/below here: https://github.com/parse-community/Parse-Swift/issues/3#issuecomment-657700606

cbaker6:
@pranjalsatija if you can get 100, that would be great, but don't feel like that "has" to be the case. For now, covering the important areas that will know will be used is most important. Based on what you mentioned, in some of those cases I think you can simply test the failure case in your test, though I never tested a fatalError directly, I assume there should be an XCTAssert that can catch it if needed.

Tests/ParseSwiftTests/ParseObjectBatchTests.swift
pranjalsatija: The ObjC SDK detects dirty keys locally, FWIW. In either case, we can go the direction you're proposing, but I want to point out the following line from the docs for `Swift.Equatable`:

> Equality implies substitutability—any two instances that compare equally can be used interchangeably in any code that depends on their values.

If we don't adhere to this, we're not following the conventions for Equatable, and that might have unintended consequences down the road.

cbaker6: I believe the parts you mention speak to the points I was making about the same objectId, they are definitely interchangeable because they have the same objectId. Once a sync happens they should all have the same data, meaning they are ==.

pranjalsatija: Right, so objects with the same data should compare equal using `==`, right? And objects with the same ID can be compared directly using their IDs. Another thing I just thought of is that if we override `==`, users will no longer be able to rely on Swift to synthesize an implementation for deep equality.

SebC99:
> @pranjalsatija a quick note about `Identifiable`, you are correct in it's usage, the problem for us is that it's iOS13 and up so it will limit us. I think comparing the objectId is okay

Why would that be an issue? Is there a discussion somewhere about the minimal iOS version this SDK should support? I guess that would be interesting. As far as I can see, when you look at the actual devices compatibility, almost every devices that are using iOS 12 can update to iOS13.
In our app, devices without that capabilities are below 1%, so it would make sense to write this SDK for modern usage if it's easier and if the performance is better doing so.
What do you think?

cbaker6:
Making this SDK modern is a fair point. One can argue if someone needs a Parse SDK that works on older devices they should use the Obj-C SDK which seems reasonable. This would allow this one to switch to iOS 13 and up.

I don't know if there's an actual performance gain unless you are talking about devices that can run iOS 13+ are automatically faster than those that can't. There are somethings that be handled in a modernized way using Combine...

There was no direct discussion I've seen about this topic except that the Xcode project was made with iOS 8+. I can see people beginning to design there apps with the Swift SDK start complaining they need a lower device target, but I can see this be less frequent than most and as I mentioned we can recommend they use the Obj-C SDK.

My last tidbit is that we currently have it working with iOS 8+ and don't require anything in iOS 13+ so far. The only two possible things I've seen that we can use (but not require) is Identifiable and Combine. I'm sure there will be others in the future, but right now the SDK is flexible across devices.

mtrezza:
> As far as I can see, when you look at the actual devices compatibility, almost every devices that are using iOS 12 can update to iOS13. In our app, devices without that capabilities are below 1%

These numbers are subjective. iOS adoption stats really depend on region, demographic and audience. We should especially consider developing countries where adoption and device distribution can be significantly different from a - say European or North American market. For example iOS <13.x adoption can be >12% in some West Asian countries.

So I agree that keeping a low minimum target should be preferred if increasing the target version would just be for coding aesthetics and not introduce any technological advancement.

SebC99:
@mtrezza true.
Apart from a reactive architecture with combine there might be no real advantage or perf improvement

cbaker6:
But, we (Parse) have two frameworks. There's nothing wrong with the obj-c SDK and it currently has more features and is battle tested. Developers needing older devices can use that one. If we decide now that the Swift one will be iOS13+ (iOS 13 seems to be a good bifurcation point) we shouldn't have trouble adopting newer features from Apple in the future.

mtrezza:
Sure, if the ObjC SDK keeps getting updated with the same features as the Swift SDK I think that's valid.

drdaz:
> But, we (Parse) have two maintained frameworks. There's nothing wrong with the obj-c SDK and it currently has more features and is battle tested. Developers needing older devices can use that one. If we decide now that the Swift one will be iOS13+ (iOS 13 seems to be a good bifurcation point) we shouldn't have trouble adopting newer features from Apple in the future.
> 
> I think we are in a unique position to design in a futuristic way without limiting the users since we have two...

I agree with this in principle. My only concern here though is that it's already hard to find maintainers for the obj-c code, and that's unlikely to improve over time. Developer time is a real bottleneck here.

Tests/ParseSwiftTests/ParseObjectBatchTests.swift
cbaker6: I'm saying it's harder to compare if two objects are actually the same and I don't think that should be the goal of ==. == should compare the same objectId because when they resolve (typically at the pare-server level) they are exactly the same. There's no way to guarantee all of the data is the same when all the data is local at the device and we shouldn't be making any promises here because of synchronization. I wouldn't try to compare the data locally, if there's a synching problem, usually apps ask the user which data do they want to keep without making a decision for them. In a centralized case, the parse-server will be asking for this resolution

cbaker6:
> I agree with this in principle. My only concern here though is that it's already hard to find maintainers for the obj-c code, and that's unlikely to improve over time. Developer time is a real bottleneck here.

True, but currently anyone who has been doing iOS Parse has to be using the Obj-C SDK (ignoring web-based apps). The Swift SDK is definitely not a drop in replacement for the Obj-c one as it's designed differently and will take time for a developer to convert. When considering that we are talking about an SDK that supports older devices, we already know that the Objc SDK works on iOS 13 (and essentially 14) and below. That should buy us a significant amount of time IMO and that's assuming Objc SDK development completely stopped today, which I don't see happening anytime soon.

Tests/ParseSwiftTests/ParseObjectBatchTests.swift
cbaker6: I'll say the big issue here is because our objects are not immutable (I'm not arguing for or against), if they were, a new objectId will be given to an object whenever a modification occurs, requiring a connection to the parse-server to get the new objectId of course. In this way, you would ensure that two objects with the same objectId actually have the same data.

drdaz:
> True, but currently anyone who has been doing iOS Parse has to be using the Obj-C SDK (ignoring web-based apps). The Swift SDK is definitely not a drop in replacement for the Obj-c one as it's designed differently and will take time for a developer to convert. When considering that we are talking about an SDK that supports older devices, we already know that the Objc SDK works on iOS 13 (and essentially 14) and below. That should buy us a significant amount of time IMO and that's assuming Objc SDK development completely stopped today, which I don't see happening anytime soon.

Agreed 🙂. Yeah it could well work out.

cbaker6:
> Sure, if the ObjC SDK keeps getting updated with the same features as the Swift SDK I think that's valid.

@mtrezza I think it may be hard to keep the features in sync. For example, the Obj-C is way ahead of Swift for now and definitely has more robust testcases. My guess is from skimming the SDKs is that some are able to leverage the latest parse-server features better than others. Doesn't the JS SDK leverage the query ability better than Obj-c? If this is true, from a query standpoint, the Swift SDK may jump ahead of the Obj-C one soon as it may be straightforward to port the JS query features. I don't think the Swift SDK should be limited by the Obj-C one in these cases

mtrezza:
> I don't think the Swift SDK should be limited by the Obj-C one in these cases

Definitely not, I agree. It depends on the feature. For example, the ObjC SDK should not become unusable for example due to iOS API changes and lack of maintenance, before we reach a significantly higher number of iOS 13 adoption. I would also say, as long as there is a workaround, such as using the JS SDK via a Cloud Code function, it's also fine, regarding your query example.

Tests/ParseSwiftTests/ParseObjectBatchTests.swift
pranjalsatija: That makes sense, but how would we implement deep equality in the Swift SDK if we replace `==`? We'd squash the implementation that the compiler synthesizes it. I just want to hear what the *advantage* of having `==` compare IDs would be.

The way I see it, there isn't one: we (and users) can very easily fall back to `hasSameObjectId`, *and* we aren't breaking the definition of Equatable. I don't think 2 objects with the same ID are equivalent if they have different data. For example, look at this situation:

```swift
class MyViewController: UIViewController {
  var gameScore: GameScore = ... // gets set on init or something

  func configureUI(with gameScore: GameScore) {
    guard gameScore != self.gameScore else { return } // to avoid repeating the same work
    // update the UI
  }
}
```

If `==` compares IDs, this (and countless other situations where users are trying to check real equality) are impossible. As for making promises about synchronization, without a way to inspect dirty keys, we are definitely lacking functionality that's currently present in the ObjC and Android SDKs, which is only sending back dirty keys to a server when an object is updated. The idea is to conserve bandwidth: if you download an object with 1000 fields and update 2 of them locally, the device only sends the 2 that have been updated back to the server. We certainly cannot guarantee that the local copy of an object we have is the same as the copy on the server, but we can and should be able to guarantee that 2 local copies do or don't have the same data.

I think it's clear at this point which approach I prefer, but I'm not unwilling to consider having `==` compare IDs; **I'd just like to know what material and concrete advantages that would provide.** There's a real concrete advantage in NOT doing anything with `==`, which is getting the compiler implementation of Equatable for free; then, there are several subjective advantages like:
* Sticking to the definition of Equatable.
* Following the conventions that every Swift struct already has; users expect their structs to compare equal using `==` if they have the same data.
* Sticking to the [Principle of Least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

Another approach would be to leave `Equatable` out of the SDK entirely and just let users decide what they want to do for their own use case. Internally, we could use `hasSameObjectId` and a similar helper for deep equality if we need it. I'm open to any of the 3 solutions, but I'd like to do the one that is the most beneficial and the least surprising for end users.

cbaker6: `hasSameObject` can work and maybe it’s okay for this to leave Equatable in tact, so I’m fine for now

For the situation you mentioned, I would expect the developer to pull the latest version of game score from the Parse-Server before displaying, refreshing self.gamescore. If they don’t have access to Internet, they may hit their cache (or local storage) for their latest local version. Using the regular wall clock scenario, they would determine the latest by sorting by updatedAt, but all of this is depending on use-case and setup...

pranjalsatija: I think our best bet is to revisit later based on feedback and what we think when we use it in apps in real life. For now, we can stick with what's there, and adapt as we move on.

pranjalsatija:
@cbaker6 so I got into writing more tests for `ParseEncoder`, and there's actually no way to catch a `fatalError` in tests; it sends a UNIX signal to the process running the test and immediately kills it. The only way to test this would be to set up a signal handler within the test suite, but I'd say that's probably not worth how cumbersome it is. I'll get the tests for non-fatal error situations done sometime soon though.

pranjalsatija:
Still working on those new tests! I actually found a couple of pretty major bugs when encoding more complex objects (thanks @cbaker6 for suggesting more tests) and I'm working on fixing them. It's been a busy week at work but I'll have them done soon.

PR #14, opened 2020-07-22
Add unit tests for API, ParseObject, and ParseUser

cbaker6:
Items added:

- [x] URLMock - this can be used for mocking any url call or error response
- [x] APICommandTests
- [x] ParseObjectCommandTests
- [x] UserCommandTests
- [x] Modernized URLSession to leverage Result type

This is ready for review...


cbaker6:
@pranjalsatija I attempted to minimize merging issues with #13. I know you made a modification to `getEncoder`, you will see that I had to slightly modify `getEncoder` by creating another that doesn't skip keys. This was done for testing purposes as I needed something that would act like the parse-server with keys like objectId, createdAt, etc that would encode. You will need to do something similar so we can mock the server in tests

Sources/ParseSwift/Objects Protocols/ObjectType.swift
```diff
@@ -46,6 +46,14 @@ internal extension ObjectType {
     func getEncoder() -> ParseEncoder {
         return getParseEncoder()
     }
+
+    func getTestDecoder() -> JSONDecoder {
+        return getDecoder()
+    }
+
+    func getEncoderWithoutSkippingKeys() -> ParseEncoder {
```
cbaker6, line 54: This is what I meant by skipping keys, a simple change that is meant to only be used inside the framework for testing. I was checking it with `getTestDecoder`, and that may be used for other tests in the future

Sources/ParseSwift/Objects Protocols/ObjectType.swift
```diff
@@ -133,11 +141,20 @@ func getJSONEncoder() -> JSONEncoder {
...
 
 func getParseEncoder() -> ParseEncoder {
+    return getParseEncoder(skipKeys: true)
+}
+
+func getParseEncoderWithoutSkippingKeys() -> ParseEncoder {
+    return getParseEncoder(skipKeys: false)
+}
+
+internal func getParseEncoder(skipKeys: Bool) -> ParseEncoder {
```
cbaker6, line 151: This was the original getParseEncoder

cbaker6:
@TomWFox @pranjalsatija when you get a chance can you look this over? It’s mostly unit tests verifying that the codebase is working as expected.It also has some minor updates

Sources/ParseSwift/Objects Protocols/ObjectType.swift
pranjalsatija: This makes sense, but is there a way to have the encoders used only for testing live in the test suite itself, instead of including them in the main codebase?

Sources/ParseSwift/Objects Protocols/ObjectType.swift
```diff
@@ -133,11 +141,20 @@ func getJSONEncoder() -> JSONEncoder {
 private let forbiddenKeys = ["createdAt", "updatedAt", "objectId", "className"]
 
 func getParseEncoder() -> ParseEncoder {
+    return getParseEncoder(skipKeys: true)
+}
+
+func getParseEncoderWithoutSkippingKeys() -> ParseEncoder {
```
pranjalsatija, line 147: Is this function necessary? Can we just call `getParseEncoder` directly and give it the `skipKeys` param?

Tests/ParseSwiftTests/ParseObjectCommandTests.swift
```diff
@@ -0,0 +1,165 @@
...
+        //: Your own properties
+        var score: Int
+
+        //: a custom initializer
+        init(score: Int) {
+            self.score = score
+        }
+    }
+
+    let parseDateEncodingStrategy: ParseEncoder.DateEncodingStrategy = .custom({ (date, enc) in
```
pranjalsatija, line 31: Does this have to be redefined here? Is it possible to use the one that's in `ParseCoding`? You might have to pull my PR into your branch for those changes to become visible, or we could maybe change this after merging?

Tests/ParseSwiftTests/ParseObjectCommandTests.swift
```diff
@@ -0,0 +1,165 @@
...
+            XCTAssertNotNil(fetched)
+            XCTAssertNotNil(fetched.createdAt)
+            XCTAssertNotNil(fetched.updatedAt)
+            XCTAssertNil(fetched.ACL)
+        } catch {
+            XCTFail(error.localizedDescription)
+        }
+    }
+
+    func testSaveCommand() {
```
pranjalsatija, line 112: Can you rename this to `testUpdateCommand`? And possibly add a separate test for creating brand new objects? This one already has an ID, but I think we should also test the case where we don't.

Tests/ParseSwiftTests/ParseUserCommandTests.swift
```diff
@@ -0,0 +1,251 @@
...
+            self.ACL = nil
+            self.customKey = "blah"
+            self.sessionToken = "myToken"
+            self.username = "hello10"
+            self.password = "world"
+            self.email = "hello@parse.com"
+        }
+    }
+
+    let parseDateEncodingStrategy: ParseEncoder.DateEncodingStrategy = .custom({ (date, enc) in
```
pranjalsatija, line 59: Ditto here about using the strategy from the main codebase.

Sources/ParseSwift/Objects Protocols/ObjectType.swift
cbaker6: This is automatically marked internal in the framework meaning the only way it can be accessed is by using `@testable import ParseSwift`

Tests/ParseSwiftTests/APICommandTests.swift
```diff
@@ -0,0 +1,138 @@
...
+                              masterKey: "masterKey",
+                              serverURL: url)
+    }
+
+    override func tearDown() {
+        super.tearDown()
+        MockURLProtocol.removeAll()
+    }
+
+    func testConstructors() {
```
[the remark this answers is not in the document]
cbaker6: True, I can change

Tests/ParseSwiftTests/ParseObjectCommandTests.swift
cbaker6: This was left over from when I was testing the JSON decoder and date. I'll remove as I'm not using it

Tests/ParseSwiftTests/ParseObjectCommandTests.swift
cbaker6: Good catch! I'll add the without objectId case and switch these two to Update

Sources/ParseSwift/Objects Protocols/ObjectType.swift
cbaker6: In it's current setup, you can call `getParseEncoder` with the `skipKeys` parameter. I added `getEncoderWithoutSkippingKeys` as an extension of `ObjectType`, `getParseEncoder` isn't part of that extension. These are things I didn't want to move around because it can easily create merge conflicts for your PR, but it's possible depending on the changes in your PR that we can do something like this after it's merged into yours

Tests/ParseSwiftTests/ParseObjectCommandTests.swift
cbaker6: Fixed, I also did this for User and made more realistic server responses to update and save

pranjalsatija: Awesome!

pranjalsatija:
LGTM! We can touch one or two things up once all the open PRs are merged.

TomWFox:
@pranjalsatija can you take another look at this and then I can merge?

cbaker6:
@pranjalsatija the only real changes are the minor ones I made to URLMock, the other SPM files I added, I deleted later as they weren't needed

cbaker6:
@TomWFox no more changes from me on this one

PR #15, opened 2020-07-23
Async API+Command

cbaker6:
Add async API, all networking calls are pushed to URLSession.shared, which uses the necessary threads carry out the networking request.

This PR adds/changes the following:

- [x] Modified Async API - previous version went async->sync->async, now it's async
- [x] Modified Sync API - previous version used URLSession directly, this one syncs the executeAsync now
- [x] Added callbackQueue to async API calls. When async calls are made, the developer can specify the queue to return the call on. It defaults to the main queue if no queue is specified
- [x] Modernized Async and batch calls so they return a Result type for better data/error handling
- [x] Bug fixes to batching
- [x] Bug fixes when saving a response from the server
- [x] Made the framework easier to test by marking some value types "internal" and "Codable" when needed
- [x] Testcases for most of the async calls*

*The test cases are on this [branch](https://github.com/netreconlab/Parse-Swift/tree/async). They will be added after #14 is merged as they are dependent on URLMocking.


mtrezza:
> There's an issue because the async session gets killed, since it's not synchronous and nothing is holding the schedule, killing the session. The current way is still blocking the calling thread, allowing the async to finish.

Can you give more details about the issue?

Sources/ParseSwift/API/URLSession+async.swift
```diff
@@ -0,0 +1,89 @@
...
+
+import Foundation
+import Combine
+
+extension URLSession {
+
+    internal func asyncDataTask(with request: URLRequest,
+                                completion: @escaping(Result<Data, ParseError>) -> Void) {
+
+        let semaphore = DispatchSemaphore(value: 0)
```
cbaker6, line 17: @mtrezza This should be commented out along with the rest of semaphore

Sources/ParseSwift/API/URLSession+async.swift
```diff
@@ -0,0 +1,89 @@
...
+            semaphore.signal()
+        }.resume()
+        semaphore.wait()
+    }
+
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
+    internal func asyncDataTask(with request: URLRequest) -> Future<Data, ParseError> {
+
+        return Future<Data, ParseError> { promise in
+            let semaphore = DispatchSemaphore(value: 0)
```
cbaker6, line 40: @mtrezza This should be commented out along with the rest of semaphore

Sources/ParseSwift/API/URLSession+async.swift
```diff
@@ -0,0 +1,89 @@
...
+                    return
+                }
+                completion(.failure(error))
+                return
+            }
+
+            completion(.success(responseData))
+            semaphore.signal()
+        }.resume()
+        semaphore.wait()
```
cbaker6, line 33: You can test if your code works by simply commenting out this line

Sources/ParseSwift/API/URLSession+async.swift
```diff
@@ -0,0 +1,89 @@
...
+
+extension URLSession {
+
+    internal func asyncDataTask(with request: URLRequest,
+                                completion: @escaping(Result<Data, ParseError>) -> Void) {
+
+        let semaphore = DispatchSemaphore(value: 0)
+        dataTask(with: request) { (responseData, urlResponse, responseError) in
+
+            guard let responseData = responseData else {
```
cbaker6, line 20: Basically nothing inside the completion fires without the semaphore because it gets killed when it goes out-of-scope.

Sources/ParseSwift/API/URLSession+async.swift
mtrezza: And what is the issue when you comment it out?

Sources/ParseSwift/API/URLSession+async.swift
mtrezza: Right, because nothing is hold on to `dataTask`. And why do you want to remove the semaphore?

Sources/ParseSwift/API/API+Commands.swift
```diff
@@ -55,6 +55,62 @@ internal extension API {
...
+                        guard let decoded = try? self.mapper(responseData) else {
+                            guard let parseError = try? getDecoder().decode(ParseError.self, from: responseData) else {
+                                completion(nil, ParseError(code: .unknownError, message: "cannot decode error"))
+                                return
+                            }
+                            completion(nil, parseError)
+                            return
+                        }
+                        completion(decoded, nil)
+                    })
```
cbaker6, line 90: With combine, you typically would add the subscription to the publisher here. If it was an instance that wasn’t killed, it would be easy and solve the problem. It’s not as easy here because it’s a value type and can’t be mutated.

Sources/ParseSwift/API/URLSession+async.swift
cbaker6: It’s blocking the thread with the semaphore there, so it isn’t truly Async. Leaving the semaphore there is the same as the sync case

Sources/ParseSwift/API/URLSession+async.swift
cbaker6: The data will never be retrieved from the pare-server. In the previous PR, I added URLMocking, running a test with it shows no data is ever retrieved

cbaker6:
@pranjalsatija let me know if you have some thoughts on how to fix this? I know you and Flo were discussing API which led to the current version.

The sync version of “async” works, so it’s possible there’s a quick fix

pranjalsatija:
I took a look, and the thing that sticks out the most to me is the fact that the async version completely reimplements the sync version. I think it would be much easier to just wrap the sync version in a call to `DispatchQueue.async` and call it a day. I have a new branch on my repo called `async_api` that you can check out if you'd like, where I've implemented those changes. Basically, I added a `queue: DispatchQueue` property to `API.Command` where each queue gets a label that's randomly generated with `UUID`. All async calls just do `queue.async` and wrap the sync versions.

cbaker6:
> I took a look, and the thing that sticks out the most to me is the fact that the async version completely reimplements the sync version. I think it would be much easier to just wrap the sync version in a call to DispatchQueue.async and call it a day. I have a new branch on my repo called async_api that you can check out if you'd like, where I've implemented those changes. Basically, I added a queue: DispatchQueue property to API.Command where each queue gets a label that's randomly generated with UUID. All async calls just do queue.async and wrap the sync versions.

I tried something similar at first, but it didn't work work. It's possible I set it up wrong. I'll take a further look at your branch and see if I can test it

cbaker6:
> I think it would be much easier to just wrap the sync version in a call to DispatchQueue.async and call it a day.

I'm still looking, but the problem with this is that `URLSession` is async by design and made in a way to not require Dispatch. The current synchronous design of execute turns the asynchronous `URLSession` into synchronous. By wrapping, we are essentially going from "async->sync->async". Removing the semaphores in the executeAsync allows us to leverage the already async `URLSession`, but it doesn't work, I simply have the semaphores there to ensure that it code works, not to reimplement the original one

cbaker6:
@pranjalsatija I'm not able to see your `async_api` branch. Did you push it too GitHub?

pranjalsatija:
Just did! Sorry about that.

cbaker6:
> Just did! Sorry about that.

@pranjalsatija I don't think your updates were committed. The last commit I see is from this branch

pranjalsatija:
🤦‍♂️ Fixed it. My bad.

cbaker6:
@pranjalsatija the changes you made has the same type of issues I mentioned on here. Also be sure to look at my note [here](https://github.com/parse-community/Parse-Swift/pull/15#issuecomment-663625761). When we finally fix the async case, we can wrap that one to make the sync case, but I don't think it should be the other way around for the reasons I stated in the comment. I left the current sync case intact as it works and I didn't want to break it, which is why I copied portions of it.

You can test your code with the test case I made in this branch https://github.com/netreconlab/Parse-Swift/tree/async_other

The test case is here: https://github.com/netreconlab/Parse-Swift/blob/9032cc370044a8e63559297c9886268829b21913/Tests/ParseSwiftTests/ParseObjectCommandTests.swift#L231

You can set a break point here: https://github.com/netreconlab/Parse-Swift/blob/9032cc370044a8e63559297c9886268829b21913/Tests/ParseSwiftTests/ParseObjectCommandTests.swift#L251

If you hit that break point without using the semaphore/wait, etc. then you made async work.

pranjalsatija:
Have you tried using expectations in the test case? I'm not on my laptop right now but a quick glance at the test code shows that there's no `waitForExpectation` which might explain why the breakpoint isn't getting hit: Xcode is terminating the test because it doesn't know it's supposed to wait for an asynchronous task to complete.

pranjalsatija:
If the test shows up in Xcode as passed but the breakpoint isn't getting hit, that's almost certainly the case.

cbaker6:
@pranjalsatija I haven't tried that, but will now, thanks!

pranjalsatija:
Sounds good!

And about your earlier comment regarding, sync to async back to sync: I agree. If the new async implementation works, we can keep it and implement the sync functionality using the async implementation.

cbaker6:
@pranjalsatija this worked for the branch I made for your code. It didn't work on my original async branch though.

My initial thinking is that this is moving the networking off of the `com.parse.ParseSwift.async` queue and into a thread labeled by the UUID you made (which behaves synchronously), but since it's off the `com.parse.ParseSwift.async`, API can still process more calls (this might be okay). If we want to serialize all of the networking calls, we can move them to a dedicated networking thread instead of the UUID to ensure they are processed in the order they are made (this would behave similar to the current `execute`. 

I will look into this more, thanks for your help!

Let me know if you have anymore ideas

pranjalsatija:
I agree, I just have one or two things to add:
* The networking is synchronous, but since each command gets its very own DispatchQueue, I don't think it'll limit concurrency or cause any other problems. Even so, if the plan is to use async networking by default and implement sync methods using the async ones, this won't matter anyway.
* I think moving all networking operations onto a single queue just for networking is a decent idea, as long as we document it. I think it'll make concurrency easier later on when we start getting into local data storage and other related functionality.

cbaker6:
> The networking is synchronous, but since each command gets its very own DispatchQueue, I don't think it'll limit concurrency or cause any other problems. 
I think you are right since all calls to the network are using `URLSession.shared...`

cbaker6:
Made changes to get this working (thanks again @pranjalsatija for the test case expectation, that was definitely the issue). `execute` now leverages `executeAsync` to reduce duplicate code. I've added all of the test cases in a different [branch](https://github.com/netreconlab/Parse-Swift/tree/async) and will bring all of the code over to this one once PR 13 is merged. After adding that code, this branch is ready for review...

A note about using `Combine` publisher/subscriber. This versions of the code doesn't use it, mainly because it looks like the subscriber gets killed when it goes out scope (the issues I was mentioning at the beginning of the thread). The publisher is still alive and fires `Future` like it's suppose to, but there's no subscriber left to receive the data. This can be looked at in the future, but isn't a big deal now is the current version is still async. All it means is the we are not using the most modern swift style code which is only available for iOS13+ anyways

cbaker6:
@pranjalsatija, @mtrezza, @drdaz when you get a chance can start looking this over?

By writing the test cases (I'll mention these in a few), I managed to discover/squash a number of bugs in the framework. Some notable ones are:

- Deleted Asynchronous.swift as the async is handled by URLSession directly
- Fixed bugs with batching
- Fixed a bug in SaveResponse in which updatedAt was the same as createdAt instead of accepting response from the server
- Moved some value types from public to internal
- Made some value types Codable instead of Decodable so they can be used in tests
- Many other minor fixes/changes

The additional files that are missing are test cases which are sitting on my [async branch](https://github.com/netreconlab/Parse-Swift/tree/async). You can clone and run the testcaes locally if you want since I'm waiting until PR 14 is merged before I add them here. The relevant testcases for this particular PR are:

- [ParseQueryTests.swift](https://github.com/netreconlab/Parse-Swift/blob/async/Tests/ParseSwiftTests/ParseQueryTests.swift)
- [ParseObjectBatchTests.swift](https://github.com/netreconlab/Parse-Swift/blob/async/Tests/ParseSwiftTests/ParseObjectBatchTests.swift)
- [ParseObjectCommandTests.swift](https://github.com/netreconlab/Parse-Swift/blob/async/Tests/ParseSwiftTests/ParseObjectCommandTests.swift) - All tests that have "Async" in the name, the others are in PR 14
- [ParseUserCommandTests.swift](https://github.com/netreconlab/Parse-Swift/blob/async/Tests/ParseSwiftTests/ParseUserCommandTests.swift) - All tests that have "Async" in the name, the others are in PR 14

cbaker6:
This is ready for review. We are up from 19 to 70+ tests, so we can catch a lot more breaking changes than before

TomWFox:
That's fantastic, I don't have time (or probably the knowledge!) to look at this in detail so I'll leave that to @pranjalsatija when he's available.

It would be great to get the tests into CI at some point - in the obj-c sdk the tests run in circle ci but you could use travis or something else if you have a preference.

cbaker6:
@TomWFox I have all the tests running in Travis during the SPM phase 🙃 https://travis-ci.org/github/parse-community/Parse-Swift/jobs/713122628

If something breaks, we will lose the green check mark. Currently Travis is only set to build for iOS and macOS, but it should be straightforward to add the others

TomWFox:
@cbaker6 I promise I looked at the config but missed that! 🤦

Sources/ParseSwift/Objects Protocols/UserType.swift
```diff
@@ -33,20 +41,48 @@ public extension UserType {
...
     static func logout() throws {
-        _ = try logoutCommand().execute(options: [])
+        _ = try logoutCommand()
+            .execute(options: [])
+    }
+
+    static func logout(callbackQueue: DispatchQueue = .main, completion: @escaping (Result<Bool, ParseError>) -> Void) {
+        logoutCommand()
+            .executeAsync(options: [], callbackQueue: callbackQueue) { result in
+                switch result {
```
pranjalsatija, line 74: The switch can be replaced with `completion(result.map { true })`

Sources/ParseSwift/Types/Query.swift
```diff
@@ -24,6 +30,15 @@ extension Querying {
     func count() throws -> Int {
         return try count(options: [])
     }
+    func find(completion: @escaping (Result<[ResultType], ParseError>) -> Void) {
```
pranjalsatija, line 33: Should this method (and also `first` and `count`) take a callback queue instead of hardcoding it to `.main`?

Tests/ParseSwiftTests/ParseObjectBatchTests.swift
```diff
@@ -0,0 +1,722 @@
...
+                        return
+                }
+                guard let originalCreatedAt = score.createdAt,
+                    let originalUpdatedAt = score.updatedAt else {
+                        XCTFail("Should unwrap dates")
+                        return
+                }
+                /*Date's are not exactly as their original because the URLMocking doesn't use the same dateEncoding
+                 strategy, so we only compare the day*/
+                XCTAssertTrue(Calendar.current.isDate(savedCreatedAt, equalTo: originalCreatedAt, toGranularity: .day))
```
pranjalsatija, line 228: Can you elaborate on this a little? Why / how are they not the same, and if we can't work around that at all, can we use a finer granularity? A day seems like a really wide range.

Sources/ParseSwift/API/API+Commands.swift
```diff
@@ -34,27 +35,60 @@ internal extension API {
...
+            group.wait()
+
+            guard let response = responseResult else {
+                throw ParseError(code: .unknownError,
+                                 message: "couldn't unrwrap server response")
+            }
+            return try response.get()
+        }
+
+        public func executeAsync(options: API.Options, callbackQueue: DispatchQueue?,
```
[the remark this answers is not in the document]
cbaker6: The callbackQueue for all of the higher API calls default to .main. For those who are adding internally to Parse-Swift, they should specify directly.

Also, the sync one shouldn’t return to the main queue as it won’t work properly, which is why it passes in nil. The sync one returns to the queue it’s called on

Sources/ParseSwift/API/API+Commands.swift
```diff
@@ -34,27 +35,60 @@ internal extension API {
...
-            } catch {
-                do {
-                    throw try getDecoder().decode(ParseError.self, from: responseData)
-                } catch {
-                    throw ParseError(code: .unknownError, message: "cannot decode response: \(error)")
+
+            URLSession.shared.dataTask(with: urlRequest, callbackQueue: callbackQueue) { result in
+                switch result {
+
+                case .success(let responseData):
```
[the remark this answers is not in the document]
cbaker6: What do you think the other way should be?

I think there’s two types of success, one is just a standard HTTP(s) success which is just being able to connect and the other is communicating to the parse server correctly

Sources/ParseSwift/API/BatchUtils.swift
```diff
@@ -46,10 +42,11 @@ struct SaveOrUpdateResponse: Decodable {
...
     }
 
-    func apply<T>(_ object: T) -> T where T: ObjectType {
-        if isCreate {
+    func apply<T>(_ object: T, method: API.Method) -> T where T: ObjectType {
+        switch method {
+        case .POST:
             return asSaveResponse().apply(object)
-        } else {
+        default:
```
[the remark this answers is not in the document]
cbaker6: This makes sense, I can make this change

Sources/ParseSwift/Types/Query.swift
cbaker6: I’ll verify this later, but If I remember correctly, these extensions are setup to create the default .main you mentioned earlier

cbaker6:
@TomWFox can you set `$GITHUB_TOKEN` In github settings for the pages deployment? I think that will fix the deployment I’m testing in the ci branch

Tests/ParseSwiftTests/ParseObjectBatchTests.swift
cbaker6: Since I’m mocking in the test, I’m attempting to create responses the server would have, but don’t have full access to how it creates things like dates. In some cases, I can just use the encoder/decoder in the framework. If you attempt to completely compare the dates directly here, the tests will fail, I believe because of the reasons I mentioned.

I don’t believe this test needs to be more granular as I specify the difference in days for the test case so it’s the only thing that needs to be compared

cbaker6: Your comment made me look back at this and I fixed the problem. The issue was `fetch` never used `apply` after it decodes so the dates wasn't being applied correctly

Sources/ParseSwift/Objects Protocols/UserType.swift
cbaker6: Good catch!

Sources/ParseSwift/API/URLSession+extensions.swift
```diff
@@ -0,0 +1,55 @@
...
+//  Original file, URLSession+async.swift, created by Florent Vilmart on 17-09-24.
+//  Name change to URLSession+extensions.swift and support for sync/async by Corey Baker on 7/25/20.
+//  Copyright © 2020 Parse Community. All rights reserved.
+//
+
+import Foundation
+
+extension URLSession {
+
+    internal func dataTask(with request: URLRequest, callbackQueue: DispatchQueue?,
```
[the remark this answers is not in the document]
cbaker6: Your implementation definitely reads better, thanks!

Sources/ParseSwift/API/API+Commands.swift
```diff
@@ -123,20 +157,34 @@ extension API.Command where T: ObjectType {
...
-            return bodies.enumerated().map({ (object) -> (T, ParseError?) in
-                let response = responses[object.0]
-                if let success = response.success {
-                    return (success.apply(object.1), nil)
-                } else {
-                    return (object.1, response.error)
+            do {
+                let responses = try getDecoder().decode(decodingType, from: data)
+                return bodies.enumerated().map({ (object) -> (Result<T, ParseError>) in
+                    let response = responses[object.0]
```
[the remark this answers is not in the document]
cbaker6: What do you think about the update?

Sources/ParseSwift/Objects Protocols/ObjectType.swift
```diff
@@ -182,10 +182,26 @@ public extension ObjectType {
         return try saveCommand().execute(options: options)
     }
 
+    func save(options: API.Options, callbackQueue: DispatchQueue,
```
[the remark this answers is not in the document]
cbaker6: Good catch, I did this on most of the higher level API calls, but missed a couple that you found

Sources/ParseSwift/Types/Query.swift
cbaker6: These are setup correctly as there's another extension in the file on Line 212 that allows for both `options` and `callbackQueue` to be used. The ones you referenced are to make it easier for the user as you pointed out earlier

Sources/ParseSwift/API/API+Commands.swift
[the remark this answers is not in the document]
pranjalsatija: It looks great!

Sources/ParseSwift/API/API+Commands.swift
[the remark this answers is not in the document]
pranjalsatija: That makes sense; what if HTTP failures AND Parse errors both fall under the failure case then?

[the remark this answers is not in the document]
cbaker6: This is a good question...

For some reason github isn’t let me post on the lines directly, but if the parse-server sends an error it will hit the completion in line 88. Any other error will hit the completion on line 85. I made sure to test parse-errors, http errors, and a JSON error in the test cases

Sources/ParseSwift/API/Responses.swift
```diff
@@ -33,3 +33,15 @@ internal struct UpdateResponse: Decodable {
         return object
     }
 }
+
+internal struct FetchResponse: Decodable {
```
cbaker6, line 37: My last commit may look like a big change, but it's mostly in the test cases. Basically I started using the getTestDecoder() I added in one of the previous commits so I can get dates in the correct format (from the dateEncodingStrategy). This allows us to compare the mocked dates to the Parse-Swift ones. Should also verify more of the ParseEncoder.

The more important part of the commit is the addition of `FetchResponse` to `SaveResponse` and `UpdateResponse`. Originally I was thinking why not just apply createdAt, updatedAt, and objectId to all objects pulled from the parse-server, but I believe I know why these are there. The business logic (createdAt, updatedAt, and objectId) is meant to all come from the parse-server and should be immutable (or almost immutable, things get strange if you updated createdAt or objectId via Parse Dashboard or some other means). The three types of responses ensure the only the correct business logic is updated locally by the API. I would imagine this is important for the local storage stuff you are working on as once the createdAt and objectId are set, they should essentially never change.

pranjalsatija: That makes sense! I think it's a good idea to defer to the server for this stuff. That being said, there's one or two things I'd like to point out:
* `SaveOrUpdateResponse` doesn't include `FetchResponse`, and `SaveOrUpdateOrFetchResponse` is a little cumbersome. So maybe `SaveOrUpdateResponse` should be renamed to `WriteResponse`?
* I don't like the signature for `apply`. I think it should be redefined so the signature reads `applied(to:)`. Since we're not mutating the object directly, it should be `applied` instead of `apply`, and the `to` just makes it read nicer. Alternatively, we could make it take the object as an inout parameter, and change it to `apply(to:)`.

Sources/ParseSwift/API/BatchUtils.swift
```diff
@@ -46,11 +42,23 @@ struct SaveOrUpdateResponse: Decodable {
         return UpdateResponse(updatedAt: updatedAt)
     }
 
-    func apply<T>(_ object: T) -> T where T: ObjectType {
-        if isCreate {
+    func asFetchResponse() -> FetchResponse {
```
cbaker6, line 45: @pranjalsatija The fetch response is here

Sources/ParseSwift/API/API+Commands.swift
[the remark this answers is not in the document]
pranjalsatija: Can you take a look at commit 07a89ca on the `async_api` branch in my fork? I updated `URLSession.dataTask` to take the mapper as one of its parameters; I think it reads better, and that logic might be more appropriate for the `URLSession` extension anyway.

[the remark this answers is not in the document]
cbaker6: I like it, I added your changes in

Sources/ParseSwift/API/Responses.swift
cbaker6: Oic, you were referring to the 'SaveOrUpdateResponse` name not including "fetch", I'm fine with `WriteResponse`. 

I made your suggested changes to "apply(to:)"

cbaker6:
@TomWFox can you approve?

drdaz:
@TomWFox can we add @cbaker6 to the objc sdk team too? If he wants to ofc.

cbaker6:
@drdaz thanks for the kudos!

I wouldn't mind helping with the obj-c project though I try not to write obj-c code nowadays, I can still help with bug fixes and CI

drdaz:
> @drdaz thanks for the kudos!
> 
> I wouldn't mind helping with the obj-c project though I try not to write obj-c code nowadays, I can still help with bug fixes and CI

I also prefer not to write obj-c. It hurts my brain and eyes way more than necessary 😃.

But I could really use an active pair of eyes over there rn.

cbaker6:
> I also prefer not to write obj-c. It hurts my brain and eyes way more than necessary 😃.

same for me!

I do use the obj-c SDK in my Swift apps heavily, so I'm definitely willing to help out

TomWFox:
@drdaz added to my mental to do list

PR #18, opened 2020-08-02
Switch to Github Actions CI and fix doc deployment

cbaker6:
Resources are shared across the parse-platform org for Circle-CI, Travis, and Github Actions. When it comes to macOS resources, each of the CI's have a much lower number of availability compared to linux builds. Basically, the objective-c uses all of Travis and Circle-CI for a very long time for it's builds.

This moves the the CI for ParseSwift to Github Actions so it doesn't contend for resources


cbaker6:
@TomWFox can you add `ACCESS_TOKEN` with `repo` access in GitHub settings? Details are here https://github.com/marketplace/actions/publish-jazzy-docs#configuration

TomWFox:
I've added a secret with the name `PERSONAL_ACCESS_TOKEN` I couldn't do `GITHUB_TOKEN` for some reason.

cbaker6:
@TomWFox the token works and the pages successfully deploy. I'll look into only getting this to push for the master branch. 

In regard to circle, it's fast, but it has a limitation of only serving 1 build for all of parse-platform. The problem with this is the obj-c build takes a long time and if we attempt to build at the same time the queue will be way too long. I'm going to migrate the rest of circle to GitHub actions (obj-c can have circle since it was there first and should use the fastest CI). Worse case I'll use a combination of actions and travis for the fastest build.

If we go with all actions, you may have to remove the travis requirement, but I'll let you know once I finish testing

cbaker6:
@TomWFox you can remove the travis requirement. Also, in circle can you "stop build" in settings?

This PR is ready for review

TomWFox:
That's all done.

TomWFox:
@cbaker6 would be nice to replace the travis badge in the readme and add a codecov one as well.

cbaker6:
@TomWFox thanks!

For some reason it's still showing circle. I think there's an extra "unfollow" you have to hit on their site

TomWFox:
Already unfollowed, it's still there. I would imagine it won't appear now on new PRs.

PR #19, opened 2020-08-22
Persist ParseUser, ParseInstallation, and default ACL in Keychain

cbaker6:
This takes a slightly different approach than its' Objective-c counterpart when storing important/private information related to the ParseUser, ParseInstallation, and default ACL. Instead of using some form of local storage such as CoreData or SQLite, it stores these items in the Keychain for safety. The Objective-c SDK does store sessionToken and some other items in the Keychain, but not whole objects as in this case.

- [x] Store ParseUser along with sessionToken in Keychain
- [x] Add ParseInstallation support and store ParseInstallation in Keychain
- [x] Fix bugs in ACL and store default ACL in Keychain
- [x] Bumped minimum OS requirements to `.iOS(.v11), .macOS(.v10_13), .tvOS(.v11), .watchOS(.v4)` due to ParseEncoder requirements. In the future may bump hire, check [here](https://github.com/parse-community/Parse-Swift/pull/13#issuecomment-669164659) for discussion.
- [x] Added [Swift Package Index](https://swiftpackageindex.com/parse-community/Parse-Swift) badges for compatibility and platforms to README
- [x] Ported some documentation from the Obj-c framework
- [x] Better organize file structure


cbaker6:
@pranjalsatija can you review this? Also, you probably know better than I would on how best to write `setDefaultACL` as I'm assuming it will use `ParseStorage.shared.use`

Tests/ParseSwiftTests/KeychainStoreTests.swift
```diff
@@ -149,20 +149,20 @@ class KeychainStoreTests: XCTestCase {
     }
 
     func testThreadSafeSet() {
-        DispatchQueue.concurrentPerform(iterations: 100) { _ in
+        DispatchQueue.concurrentPerform(iterations: 50) { _ in
```
pranjalsatija, line 152: Is there a reason these are getting turned down?

Tests/ParseSwiftTests/ParseObjectBatchTests.swift
```diff
@@ -665,7 +665,7 @@ class ParseObjectBatchTests: XCTestCase { // swiftlint:disable:this type_body_le
             }
 
         }
-        wait(for: [expectation1, expectation2], timeout: 10.0)
+        wait(for: [expectation1, expectation2], timeout: 30.0)
```
pranjalsatija, line 668: Same here; is there a reason this timeout is going up?

cbaker6: This gives more time if it's needed on slower machines. If it's fulfilled earlier it won't hit this timeout

Tests/ParseSwiftTests/KeychainStoreTests.swift
cbaker6: I turned this down originally because it would randomly fail on some tests with 100 threads. I think I fixed the problem with moving the fulfills

cbaker6:
@pranjalsatija can you review this? The large changes in lines are due to documentation and restructuring.

Most of the updates are Installation (new), ParseUser (storing to Keychain and memory), and default ACL to Keychain, along with their respective test cases

cbaker6:
@TomWFox a significant part of this PR is the API documentation since it essentially had no documentation before this. I ported the markdown comments from the obj-c framework to this one and attempted to tailor to this framework. This also switches to the "swift" style markdown (`- parameter`) from obj-c (`@param`).  The jazzy docs should have no issues with the change.

I'm sure there are number of grammatical mistakes along with better documentation needed. Feel free to make any adjustments you think are necessary.

TomWFox:
@cbaker6 Spell checked and some minor improvements.

pranjalsatija:
@cbaker6 sorry for taking so long; I noticed you were still slowly adding stuff so I figured I'd wait for at least a little while and then I got busy with school stuff. My main concern with this PR is the fact that it's directly using the Keychain from within the framework.

IIRC, Flo's idea was to make it so that users could "plug and play" different adaptors for different tasks, including the persistence of the current user, installation, etc. which is why I added `ParseStorage.use`; when users are initializing the SDK, they provide a storage provider and the SDK uses that storage provider to handle all its storage needs; if that's the case, then all storage work should go through `ParseStorage.shared`. If the user doesn't provide a storage provider, `ParseStorage.shared` should do it for them, or we should treat that situation as an unrecoverable error. Previously, I had it default to using an in memory Codable store and log a warning for good measure. If we don't want to use that approach anymore, that's also OK, I just wanted to point it out.

Other than that, I don't have any issues -- LGTM. I'll do another scan for typos and stuff.

Sources/ParseSwift/ParseConstants.swift
```diff
@@ -0,0 +1,21 @@
...
+//  ParseConstants.swift
+//  ParseSwift
+//
+//  Created by Corey Baker on 9/7/20.
+//  Copyright © 2020 Parse Community. All rights reserved.
+//
+
+import Foundation
+
+let kParseVersion = "0.0.1"
```
pranjalsatija, line 11: The `k*` constants feel out of place; can these be static members on a `Constants` enum or something?

Sources/ParseSwift/Storage/PrimitiveObjectStore.swift
```diff
@@ -39,15 +44,27 @@ struct CodableInMemoryPrimitiveObjectStore: PrimitiveObjectStore {
 
 // MARK: KeychainStore + PrimitiveObjectStore
 extension KeychainStore: PrimitiveObjectStore {
+
     func delete(valueFor key: String) throws {
-        _ = removeObject(forKey: key)
+        if !removeObject(forKey: key) {
+            throw ParseError(code: .objectNotFound, message: "Object for key \"\(key)\" not found in Keychain")
```
pranjalsatija, line 50: Do we need to throw here? IMO, if someone wants to delete something that doesn't exist, it's not a big deal. If the delete "succeeds" anyway, the net outcome is the same: the thing they're deleting doesn't exist.

pranjalsatija: By the way, good catch on the rest of the error handling; I have no clue how I missed that.

pranjalsatija (approved):

cbaker6:
@pranjalsatija thanks for reviewing!

> IIRC, Flo's idea was to make it so that users could "plug and play" different adaptors for different tasks, including the persistence of the current user, installation, etc. which is why I added ParseStorage.use; when users are initializing the SDK, they provide a storage provider and the SDK uses that storage provider to handle all its storage needs; if that's the case, then all storage work should go through ParseStorage.shared. If the user doesn't provide a storage provider, ParseStorage.shared should do it for them, or we should treat that situation as an unrecoverable error. Previously, I had it default to using an in memory Codable store and log a warning for good measure. If we don't want to use that approach anymore, that's also OK, I just wanted to point it out.

I agree with you for the most part here. With the changes, I don't intend to store all or even most of the objects in the Keychain. The only items I store are the BaseParseUser, BaseInstallation, and the default ParseACL. This not only keeps these items safe, but also provides the local storage needed, which I don't think much else will need to be stored in Keychain. This differs slightly from the objective-c framework which just stored the sessionToken in the Keychain. My thinking is keep the user and Parse required items in a safe space (Keychain) and 3rd party storage can be used for everything else via a protocol and ParseStore.shared that you made

Sources/ParseSwift/Storage/PrimitiveObjectStore.swift
cbaker6: I believe I added the throw just because it conformed. I thought more of it if there was a problem deleting and object to throw.

pranjalsatija:
That's sensible! What if I update `ParseStorage.shared` a little bit then? I know we had a prior conversation about it taking multiple kinds of stores for different tasks. Right now, it takes a single `PrimitiveObjectStore`. What if I update it to take multiple? I'm thinking something like this:

```swift
struct ParseStorage {
    struct StorageConfig {
        let keyValueStore: PrimitiveObjectStore // defaults to in memory
        let secureStore: PrimitiveObjectStore // defaults to Keychain
    }
    
    mutating func use(_ config: StorageConfig) {
        self.backingStore = store
    }
}
```

And then I can also update the parts of this PR that reference the Keychain directly to reference the secure store?

Sources/ParseSwift/Storage/PrimitiveObjectStore.swift
pranjalsatija: That makes sense; I forgot all the Keychain methods just return `Bool`.

cbaker6:
> What if I update it to take multiple?

Sounds reasonable to me. It would make it more modular. My only concern is, do we want to let the user/developer to keep/choose the `secureStore` or should we just automatically keep this info in Keychain? Also, adding it your way would probably enable the framework to run on Linux as I don't think Keychain is in Foundation for Swift Linux

cbaker6:
If you want, it might be easier to merge this PR before you make the storage changes you mentioned

pranjalsatija:
@cbaker6 I'll definitely merge this PR first! And I think letting them customize it is the way to go; you already pointed out the Linux situation and we / users might want to swap in an in-memory store for unit tests or something.

Sources/ParseSwift/ParseConstants.swift
cbaker6: Just added the enum. What do you think about this?

cbaker6:
@TomWFox can you approve this? @pranjalsatija has reviewed it

Sources/ParseSwift/ParseConstants.swift
```diff
@@ -8,14 +8,6 @@
...
-let kPFDeviceType = "ios"
-#elseif os(macOS)
-let kPFDeviceType = "osx"
-#elseif os(tvOS)
-let kPFDeviceType = "tvos"
-#elseif os(watchOS)
-let kParseDeviceType = "applewatch"
-#endif
+enum ParseConstants: String {
+    case parseVersion = "0.0.1"
```
pranjalsatija, line 12: You can make this `static let parseVersion = "0.0.1"` and then you won't have to do `rawValue`. It's a small change but kinda nice.

Sources/ParseSwift/ParseConstants.swift
```diff
@@ -8,14 +8,6 @@
 
 import Foundation
 
-let kParseVersion = "0.0.1"
-
-#if os(iOS)
-let kPFDeviceType = "ios"
```
pranjalsatija: Did these have to be deleted? You could move this inside `ParseConstants`:
```
enum ParseConstants {
  #if os(iOS)
  static let deviceType = "ios"
  #elseif os(macOS)
  ...
}
```

pranjalsatija (approved):

Sources/ParseSwift/ParseConstants.swift
cbaker6: I tried something similar but the compiler didn't like it. Since this wasn't needed anywhere else, I just moved it directly to where it's used.

Of course, if you now of a way to add it, you can make the change in the PR for the storage updates

Sources/ParseSwift/ParseConstants.swift
cbaker6: I didn't know this, I'll try it

Sources/ParseSwift/ParseConstants.swift
cbaker6: I see, with the static way it works

TomWFox (approved):

PR #21, opened 2020-09-14
Added Pointer & Geopoint support, added more features to Query

cbaker6:
- [x] Added Pointers and deep save of child objects
- [x] Added full GeoPoint support along with querying based on a geopoint or proximity. Also added test cases and documentation
- [x] Added more capabilities to Query. The Swift SDK can now do almost everything the JS SDK can do in terms of querying. Support for aggregate is still missing
- [x] Added GeoPoint demo to playgrounds
- [x] Bump Xcode environment in CI from Xcode 11.6 to Xcode 11.7
- [x] Switched ParseEncoder which is a rip and mod of the JSONEncoder from open-source Swift 5.3


cbaker6:

cbaker6:
@TomWFox I added a decent amount of documentation here, can you look it over and make any necessary changes you see?

Note that in some places, I opted to make Query and GeoPoint look more like the JS SDK instead of the obj-c one. This framework probably has more querying ability than the obj-c one now, but is still slightly behind the JS one, but I'll fix that in a different PR. Also, because of the "typeness" of Swift, I couldn't think of a good way to return standard JSON when using `explain` and `hint` when using the methods that were already in place (at least without breaking and making it look ugly). Instead, I opted to make find, first, and count have an "explainable and hintable" methods

cbaker6:
Finished adding tests for this. No more changes from me unless I catch a bug somewhere

cbaker6:
@pranjalsatija you may have already known this with the changes you made to PareEncoder, but we should note that `ParseEncoder` can't be used on `Query`. This is because of calls to `superEncoder` in `QueryWhere`. This may not be a problem as the `JSONEncoder` works for all of the query tests. I currently can't think of a need of ParseEncoder on Query anyways...

Tests/ParseSwiftTests/ParseObjectTests.swift
```diff
@@ -47,6 +77,18 @@ class ParseObjectCommandTests: XCTestCase { // swiftlint:disable:this type_body_
         try? ParseStorage.shared.deleteAll()
     }
 
+    func testObjectCreationViaStruct() {
+        XCTAssertEqual(GameScore.className, "GameScore")
+        XCTAssertEqual(GameScore(score: 10).className, "GameScore")
+    }
+
+    func testObjectCreationViaClass() {
```
pranjalsatija, line 85: Is this necessary? This feels more like it's testing Swift than our code.

Sources/ParseSwift/Parse Types/GeoPoint.swift
```diff
@@ -1,15 +1,168 @@
 import Foundation
+import CoreLocation
```
pranjalsatija, line 2: This will break compatibility with non-Apple platforms like Linux.

Sources/ParseSwift/Parse Types/GeoPoint.swift
```diff
@@ -1,15 +1,168 @@
...
-        self.longitude = longitude
+        self._latitude = latitude
+        self._longitude = longitude
+    }
+
+    /**
+      Creates a new `GeoPoint` instance for the given `CLLocation`, set to the location's coordinates.
+       - parameter location: Instace of `CLLocation`, with set latitude and longitude.
+     */
+    public init(location: CLLocation) {
```
pranjalsatija, line 75: This should be put in an `available` enclosure or defined another way. I think the best thing is make a protocol called `_CLLocationShim` or something like that and extend `CLLocation` to conform.

Sources/ParseSwift/Parse Types/Internal/ParseLocationManager.swift
```diff
@@ -0,0 +1,97 @@
...
+ ParseLocationManager is an internal class which wraps a CLLocationManager and
+ returns an updated CLLocation via the provided block.
+ When -addCurrentLocation is called, the CLLocationManager's
+ -startUpdatingLocations is called, and upon CLLocationManagerDelegate callback
+ (either success or failure), any handlers that were passed to this class will
+ be called _once_ with the updated location, then removed. The CLLocationManager
+ stopsUpdatingLocation upon a single failure or success case, so that the next
+ location request is guaranteed a speedily returned CLLocation.
+ */
+class ParseLocationMananger: NSObject {
```
pranjalsatija, line 25: Is this functionality that we can / should be providing in this framework? It won't work on non-Apple platforms and it seems like a small enough thing that mostly everyone defines their own / uses Core Location directly.

Tests/ParseSwiftTests/ACLTests.swift
```diff
@@ -5,7 +5,7 @@
 //  Created by Corey Baker on 8/22/20.
 //  Copyright © 2020 Parse Community. All rights reserved.
 //
-
+#if !os(watchOS)
```
pranjalsatija, line 8: Is this necessary to include in the tests? Are we running tests separately just for the Watch?

pranjalsatija (changes requested):

Tests/ParseSwiftTests/ACLTests.swift
cbaker6: From what I've been finding online, XCTest doesn't run on watchOS. Currently, we don't get the watchOS badge on the readme because watchOS can't compile any of the tests. I used a solution discussed on the swift repo which use `!os(watchOS)`, there's been some discussions about this, doesn't seem like there's a preferred way to go around this.

With the above flag, the readme should auto populate the watchOS badge, assuming the build goes fine

pranjalsatija:
Sure, that makes sense.

Sources/ParseSwift/Parse Types/Internal/ParseLocationManager.swift
cbaker6: I ripped this and modified it from the objc SDK. That being said, it's not necessarily a requirement, as developers will just need to make an extra step in iOS to setup core location within their app and pass them to GeoPoint, which I'm okay with.

Do you think it should be removed?

Tests/ParseSwiftTests/ACLTests.swift
pranjalsatija: Sure, that makes sense.

Sources/ParseSwift/Parse Types/Internal/ParseLocationManager.swift
pranjalsatija: My main concern is that if we include it, we have to make sure it works, and it falls outside the core responsibility of this SDK. Plus, it's a pretty simple wrapper and we're not adding / extending Core Location in any big way.

Tests/ParseSwiftTests/ParseObjectTests.swift
cbaker6: I added these tests because I ran into an issue when I was using the framework within an app.

You are right, the test isn't needed.

We should add a note in the documentation that the `ParseObject` protocol should only be used with `structs` or `final classes` as in the `class` scenario if you subclass, you will only get the base class name as the name of the ParseObject table. This differs from the objc framework of course, so users transitioning from that one may not realize this

Sources/ParseSwift/Parse Types/Internal/ParseLocationManager.swift
cbaker6: I agree, I will remove it

Sources/ParseSwift/Parse Types/GeoPoint.swift
cbaker6: I'll remove this as well and let the user user just pass long and lat

Sources/ParseSwift/Parse Types/GeoPoint.swift
cbaker6: Will remove as well

cbaker6:
@pranjalsatija It looks like the ParseEncoder in it's current state can't be used because it's missing `superEncoder`. You mentioned it's not supported in your previous PR and I think you explained why, but I couldn't find it in the discussion. Do you remember why you said it's not implemented? Did it work for the previous ParseEncoder? Basically, much of the SDK will be limited if we don't have access to superEncoder. If we can't find a way to support superEncoder, and the previous encoder does, we may have to revert back

cbaker6:
@pranjalsatija I added the superEncoder. Turns out there was a bug in the nestedContainer that prevented anything nested from encoding properly. I wrote a simple test and compared the output to the JSONEncoder (since I know that one works with the Parse-server). You can probably write more of the ParseEncoder tests in a similar way.

cbaker6:
@pranjalsatija It looks like there are number of bugs in ParseEncoder. I attempted to fix some more, but it's still broken. The easiest way to test it is to start moving the tests in the `ParseQueryTests` that use JSON encoder to ParseEncoder and watch what breaks. The JSON test cases are correct as I've tested them on a parse-server and got their dictionary counterparts from the obj-c SDK.

The `bool` issue I mentioned is another problem. For the small test cases where I got the containers to "look" correctly, the parse-server complained about this:

```
error: Parse error:  error: invalid input syntax for type double precision: "{"$gt":100}" {"code":1,"stack":"Error: error: invalid input syntax for type double precision: \"{\"$gt\":100}\"\n    at /parse-server/lib/Controllers/DatabaseController.js:1170:21\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (internal/process/task_queues.js:97:5)"}
```

It's doing something weird with conversions.

As of now, the JSONEncoder has to be used instead.

cbaker6:
Codecov dropped in this PR due to the addition of the new ParseEncoder. It's basically a mod of the native JSON encoder, so the components that are unit tested are the added Parse-Swift ones.

cbaker6:
I managed to use Parse-Swift to replace the obj-c framework in one of my sample apps where objc-parse was already working great. This is how found most of the bugs.

This framework can definitely handle the basics after this PR is merged

cbaker6:
@TomWFox I'm finished with changes/debugging. Can you review the added documentation?

TomWFox (approved):
Documentation proofread ✅

TomWFox:
Btw you can't merge this because codecov/patch is a required check but I could remove this requirement.

cbaker6:
@TomWFox i can lower the reqs for codecov patch in the .codecov file (I think) and then increase it on the next PR.

TomWFox:
Sounds good.

PR #22, opened 2020-09-15
remove some direct references to the Keychain

pranjalsatija:
@cbaker6 I think I got them all; please let me know if you see that I missed any.


cbaker6:
It looks like it broke some of the tests

pranjalsatija:
Sorry, I forgot to mark this whole thing as WIP; I wasn't fully done updating the tests!

Sources/ParseSwift/Objects/ParseUser.swift
```diff
@@ -41,27 +41,22 @@ struct CurrentUserContainer<T: ParseUser>: Codable {
...
     static var currentUserContainer: CurrentUserContainer<Self>? {
-        get {
-            guard let currentUserInMemory: CurrentUserContainer<Self>
-                = try? ParseStorage.shared.get(valueFor: ParseStorage.Keys.currentUser) else {
-                return try? KeychainStore.shared.get(valueFor: ParseStorage.Keys.currentUser)
-            }
-            return currentUserInMemory
-        }
-        set { try? ParseStorage.shared.set(newValue, for: ParseStorage.Keys.currentUser) }
+        get { try? ParseStorage.shared.secureStore.get(valueFor: ParseStorage.Keys.currentUser) }
```
cbaker6, line 44: I think this may change the original intent. Basically in this manner, any change the user makes to currentUser is automatically saved to the secureStore. Before this PR, all changes are made in memory and are only stored to the secureStore (Keychain) if the user calls .save or a variation of it.

It seems like there's no way for the user to revert a change if they accidentally changed a value, and that same changed would have to sync to the parse-server at some point.

Sources/ParseSwift/Objects/ParseUser.swift
```diff
@@ -41,27 +41,22 @@ struct CurrentUserContainer<T: ParseUser>: Codable {
...
+        set { try? ParseStorage.shared.secureStore.set(newValue, for: ParseStorage.Keys.currentUser)}
     }
 
-    internal static func saveCurrentContainerToKeychain() {
-        //Only save the BaseParseUser to keep Keychain footprint finite
-        guard let currentUserInMemory: CurrentUserContainer<BaseParseUser>
-            = try? ParseStorage.shared.get(valueFor: ParseStorage.Keys.currentUser) else {
+    internal static func saveCurrentUserContainer() {
+        //Only save the BaseParseUser to keep memory footprint finite
+        guard let currentUser: CurrentUserContainer<BaseParseUser> =
```
cbaker6, line 50: This is similar to my previous comment, basically all changes are permanent before they are actually saved to the parse-server with no way to revert

cbaker6 (changes requested):

Sources/ParseSwift/Objects/ParseUser.swift
pranjalsatija: I just want to clarify some stuff! I wasn't quite sure what the intent of the old code was, so I'll double check with you. The old code seems like it first tries to get the current installation from `ParseStorage.shared` and then tries to get it from the keychain, and if *both* fail, then it creates a new one.

The old implementation was relying on the fact that `ParseStorage.shared` is an in-memory store, which it shouldn't do: it's 100% possible that `ParseStorage.shared` (which is now `ParseStorage.primitiveStore`) is backed by the keychain, a file on disk, etc. That being said, I updated it thinking that the fallback to `KeychainStore` is only there in the event that the user didn't provide a shared store, which is no longer possible because the shared store isn't optional anymore.

As far as reverting changes goes, I don't think it's correct to say that the old version provided a way to revert changes. It just so happens to be the case that the store it was reading / writing from first was in memory, but there was (and still is) nothing enforcing that that's true.

cbaker6: > I just want to clarify some stuff! I wasn't quite sure what the intent of the old code was, so I'll double check with you. The old code seems like it first tries to get the current installation from ParseStorage.shared and then tries to get it from the keychain, and if both fail, then it creates a new one.

This is true.

> The old implementation was relying on the fact that ParseStorage.shared is an in-memory store, which it shouldn't do: it's 100% possible that ParseStorage.shared (which is now ParseStorage.primitiveStore) is backed by the keychain, a file on disk, etc. That being said, I updated it thinking that the fallback to KeychainStore is only there in the event that the user didn't provide a shared store, which is no longer possible because the shared store isn't optional anymore.

Kind of true... There wasn’t any persisted storage implemented. I used the in-memory ParseStore here on purpose so that changes to currentUser and currentInstallation happen in memory. When a “save” is successful, only then is it saved to the Keychain. This is done because there is currently no way to mark something is dirty and to determine when/how to sync with the parse-server. So the only way to confirm a change is as valid is after it’s committed to the parse-server.

> As far as reverting changes goes, I don't think it's correct to say that the old version provided a way to revert changes. It just so happens to be the case that the store it was reading / writing from first was in memory, but there was (and still is) nothing enforcing that that's true. If a user updates the value, I think we should store it persistently immediately. AFAIK, The old SDK and other SDKs like Firebase doesn't support this either; we can't be liable if the user accidentally sets the wrong value, and they should take care to write code that doesn't set the wrong value for the current user / installation.

The old version provides a way to revert because only saves to the parse-server are committed to the Keychain. If the user exits the app and iOS clears the memory, the last saved version that’s in the keychain (which was last synced to the parse-server) are returned. Memory is only committed after a save. In addition, if a user adds “bad” values to memory, they can simply pull the last version from the keychain (reverting). My point here, if there’s currently no way to tell if something is dirty, we shouldn’t be storing the data locally (which is what your changes will allow). It will make everything to get out sync. If a “bad” value is saved to the parse-server, then a “bad” value is stored in the Keychain

cbaker6: It might have gotten lost in my responses, but I believe changes in the local storage should only be allowed when there are proper mechanisms in place to sync them to the parse-server. Until then, we shouldn’t persist changes locally until the parse-server says they have been committed.

cbaker6: > The old SDK and other SDKs like Firebase doesn't support this either; we can't be liable if the user accidentally sets the wrong value, and they should take care to write code that doesn't set the wrong value for the current user / installation.

For the current version of the implementation. I didn’t take hints from the Obj-c SDK and I have no clue how Firebase works. In this situation, the local storage isn’t complete yet, so I’m not sure how they can be compared. If you have a SDK/app that relies on a centralized server (parse-server in this case), and you don’t have local storage, you should only commit after the centralized server tells you to. The obj-c has some form of local storage, so it doesn’t apply to this scenario

pranjalsatija: Then what if we update `BaseParseUser` and `BaseParseInstallation` to include an `isConfirmed` field or something that only gets set to `true` when the server has confirmed a change? And regardless of that result, until we properly implement local storage, should I make the setters internal only?

cbaker6: > Then what if we update BaseParseUser and BaseParseInstallation to include an isConfirmed field or something that only gets set to true when the server has confirmed a change?

This lets you know something has been changed locally, but how does it help synchronize the changes with the parse-server? I think syncing local changes need to be fully flushed out before allowing the user to change data locally. It will cause issues, for example, are the local changes older or newer than the parse-server? How do you know? In the current form (without full support of local storage and sync), the SDK works, with the limitation that data isn't changed until it's committed to the parse-server. This okay IMO for now until the completion of local storage.

> And regardless of that result, until we properly implement local storage, should I make the setters internal only?

Sounds reasonable to me. The one thing I've been doing is using the playground files to make sure everything is correct on the parse-server. The unit-tests are good, but they are mocking the parse-server. If you aren't already, you should make sure your changes work via playgrounds with a real parse-server. I posted some basic instructions and docker container (has parse-dashboard also) just for this [here](https://github.com/parse-community/Parse-Swift/pull/12#issuecomment-656918156).

PR #23, opened 2020-09-21
Update ACL and Xcode Project

cbaker6:
Minor updates:

- [x] Switch ParseACL to ACL to match the rest of project
- [x] Update Xcode project for Xcode 12
- [x] Add ParsePointerTests.swift
- [x] Make changes in test files to build on devices that can import XCTest
- [x] More fixes to enable building on Linux


cbaker6:
@pranjalsatija can you review? The biggest change here is ACL.

I plan on adding the File support in a different PR.

Sources/ParseSwift/Coding/AnyEncodable.swift
```diff
@@ -109,6 +109,8 @@ extension _AnyEncodable {
             throw EncodingError.invalidValue(self.value, context)
         }
     }
+
+    #if !os(Linux)
     private func encode(nsnumber: NSNumber, into container: inout SingleValueEncodingContainer) throws { // swiftlint:disable:this cyclomatic_complexity line_length
```
pranjalsatija, line 114: Why does the encoder need special behavior for Linux vs. non-Linux platforms?

Sources/ParseSwift/Object Protocols/ParseInstallation.swift
```diff
@@ -217,6 +217,7 @@ extension ParseInstallation {
             return
         }
 
+        #if !os(Linux)
```
pranjalsatija, line 220: Same here; and the next part isn't in this PR, but are you sure the desired behavior is for the Mac Catalyst bundle ID prefix to be dropped? Unless the ObjC SDK is doing the same thing, my opinion is that if Apple thinks the Catalyst app is different enough to warrant a new bundle ID, the SDK should probably respect that.

Sources/ParseSwift/Parse Types/File.swift
```diff
@@ -1,12 +1,31 @@
...
+     */
     public var data: Data?
+
+    /**
+     The url of the file.
+     */
     public var url: URL?
 
-    public init(data: Data?, url: URL?) {
+    internal init(data: Data?, url: URL?) {
```
pranjalsatija, line 28: Should this be internal?

pranjalsatija (approved):
I just have some questions / comments, but otherwise LGTM.

Sources/ParseSwift/Coding/AnyEncodable.swift
cbaker6: Good question, Swift Linux doesn't like somethings from objective-c or is missing the bridging for them, such as `NSNumber`. These I don't know specifically while coding, I figure them out by looking at the Linux build for Parse-Swift on [Swift Package Index](https://swiftpackageindex.com/builds/1160B83E-99C1-44B6-8E0E-85FD8C5833EC) and searching online. This will bring the linux build closer to fruition, but I'm sure there are still some minor issues similar to this.

Sources/ParseSwift/Parse Types/File.swift
cbaker6: There are 2 other initializers that get used publicly, one takes in data, the other takes in url (I'm adding them to the File PR later). The init for data and url are used internally in the objective-c SDK

Sources/ParseSwift/Object Protocols/ParseInstallation.swift
cbaker6: For the part, same as above.

In reference to Catalyst, I did the exact same fix the obj-c SDK does here:

https://github.com/parse-community/Parse-SDK-iOS-OSX/blob/d0119bd7602d4b880b0038b2ac2b9246044b8564/Parse/Parse/PFInstallation.m#L302-L316

Sources/ParseSwift/Coding/AnyEncodable.swift
cbaker6: I do the same for the other builds as Swift Package Index runs the unit tests for all of the builds (and then populates the badges on the Parse-Swift readme). For GitHub Actions, the test cases are only ran for iOS and SPM macOS. Builds are ran for the others in GitHub Actions, but not the unit tests.

PR #24, opened 2020-10-25
Switch ACL to ParseACL

cbaker6:
Looks like the compiler can't use the ACL as a type when the property type of an Object is named ACL. So going back to ParseACL to differentiate.


cbaker6:
@pranjalsatija this is reverting back to ParseACL because of what I mentioned above. I added a simple test case that sets defaultACL. If ParseACL is ever converted back to ACL, that particular test case won't compile

PR #25, opened 2020-11-01
Improve CI by adding macOS Xcode build

cbaker6:
Adds testing of the Xcode project macOS framework. Also, the addition of .spi.yml allows the unit tests for watchOS to be ignored on [SwiftPackageIndex](https://swiftpackageindex.com/parse-community/Parse-Swift) and we should gain the watchOS badge.


cbaker6:
@TomWFox are you able to remove "build-test-swift", "swift-macos", "swift-tvos", and "swift-watchos" from the branch protection?

You can add "xcode-test-all" and "spm-build-test". xcode-test-all replaces the other builds, but also runs the tests for them (except for watchOS which doesn't support unit testing). "build-test-swift" was renamed "spm-build-test".

cbaker6:
@TomWFox I might need to make a change, no need to remove the branch protections yet

cbaker6:
@TomWFox now it's ready. Are you able to remove "build-test-swift" and replace it with "spm-build-test"" in the branch protections?

cbaker6:
@pranjalsatija can you review this? Most of it is fixing watchOS build to ignore testing because XCTest isn't available.

PR #27, opened 2020-11-06
Fix callbackqueue bug

cbaker6:
There's a bug where the specified callback queue was being used in the incorrect place causing crashing in apps.

This most likely fixes the random crash from the tests that was occurring. This PR specifically addresses the following:
- [x] Fixes callback queue (mentioned above)
- [x] Adds tvOS testing to CI (most of the added files are due to adding TestHostTV app and it's related files)
- [x] Fixes scheme in .spi.yml so SwiftPackageIndex ignores watchOS tests


cbaker6:

cbaker6:
@TomWFox can you replace "xcode-build-tvos" with "xcode-test-tvos" in the branch protections? I've added testing to tvOS.

pranjalsatija (approved):

PR #29, opened 2020-11-10
Fix bug in nestedUnkeyedController for ParseObjects

cbaker6:
Was missing encoding children for nestedUnkeyedController


cbaker6:
@pranjalsatija Can you review this? It was a miss by me when I switched the `ParseEncoder`. Basically, when a `ParseObject` contains an array of additional ParseObjects (children), the children should be saved and converted to `Pointers`. The `ensureDeepSave` in `ParseObject` which was already there, ensures saves and has tests.

PR #30, opened 2020-11-22
Update docs and fix parse image

cbaker6:
Using the image as a link should fix it not showing up in the docs


README.md
```diff
@@ -1,5 +1,5 @@
 <p align="center">
-  <a href="https://parseplatform.org"><img alt="Parse Platform" src="https://github.com/parse-community/Parse-Swift/blob/main/logo%20large.png" width="200"></a>
+  <a href="https://parseplatform.org"><img alt="Parse Platform" src="https://user-images.githubusercontent.com/8621344/99892392-6f32dc80-2c42-11eb-8c32-db0fa4a66a81.png" width="200"></a>
```
TomWFox, line 2: Where is this link actually coming from?

dplewis fixed a [similar problem](https://github.com/parse-community/Parse-SDK-JS/pull/1236/files) with a link like this...
`https://github.com/parse-community/Parse-SDK-JS/raw/master/.github/parse-logo.png` - I'm guessing having `/raw` is what does it.

cbaker6: This one is coming from GitHub itself. I copied the original image into a GitHub issue in which GitHub automatically hosts the file, after we can use the link. From what I’ve read, GitHub doesn’t delete these files. This makes it so we don’t have to add the file to the repo.

If you wanted, all other repos can point to this same link and the file can be removed from that repo

TomWFox: Interesting, I guess it reduces the download by 27.6 KB!

TomWFox (approved):

PR #32, opened 2020-12-10
Fix password saving to Keychain after user login

cbaker6:
- [x] This makes the login behave similar to signup. There's a bug where the password isn't stored to current user after login. The local user is updated from the parse-server during signup/login. A modification was also made to tests to make LoginResponse respond like a real parse-server by removing password from the response.
- [x] ParseObjects now receive updated ACL from the server after a save 
- [x] In addition, all build warnings are fixed and the inclusive_language warnings from swiftLint are silenced until changes are made on the parse-server
- [x] swiftLint now autocorrects during builds

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -5,7 +5,7 @@ coverage:
   status:
     patch:
       default:
-        target: 0
+        target: auto
     changes: false
     project:
       default:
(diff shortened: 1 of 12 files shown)


cbaker6:

PR #33, opened 2020-12-10
Fix encoding object arrays

cbaker6:
- [x] Fix bug in custom encoder when persisting child objects
- [x] Adds back in ParseEncoder warning from original JSONEncoder (removed in #32). Something is happening that the compiler is missing and this can cause in app crash during encoding. I'm assuming the original JSONEncoder has this for a reason so we will keep for now until the native JSONEncoder changes

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -5,7 +5,7 @@ coverage:
   status:
     patch:
       default:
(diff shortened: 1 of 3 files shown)


cbaker6:

PR #38, opened 2020-12-20
Update keychains of ParseUser and ParseInstallation

### LINKED ISSUE #37: User.current not returning all variables
When a user is logged in, and you request a custom key from the user, it returns `nil` even though there is a value for that user in the database.

**This is my ParseUser structure**
```
struct User: ParseUser {
            //: These are required for ParseObject
            var objectId: String?
            var createdAt: Date?
            var updatedAt: Date?
            var ACL: ParseACL?

            //: These are required for ParseUser
            var username: String?
            var email: String?
            var emailVerified: Bool?
         
            //: Custom
            var userId: String?
 }
```

**This is what I call on launch**
```
let currentUser = User.current
        print("currentUser - objectId: \(String(describing: currentUser?.objectId))")
        print("currentUser - createdAt: \(String(describing: currentUser?.createdAt))")
        print("currentUser - updatedAt: \(String(describing: currentUser?.updatedAt))")
        print("currentUser - ACL: \(String(describing: currentUser?.ACL))")
        
        print("currentUser - username: \(String(describing: currentUser?.username))")
        print("currentUser - email: \(String(describing: currentUser?.email))")
        print("currentUser - emailVerified: \(String(describing: currentUser?.emailVerified))")
        //Custom key
        print("currentUser - userId: \(String(describing: currentUser?.userId))")
```

**This is the log**
Schermafbeelding 2020-12-20 om 00 40 06

**This is an example of 3 users in Parse database**
Schermafbeelding 2020-12-20 om 00 46 07

cbaker6:
In order to get this, User and Installation protocols have to implement it's own versions of some of the ParseObject methods. Not necessarily a problem, but adds more code to the `ParseUser` and `ParseInstallation` protocols. Also requires some additional testing outside of the `ParseObject` tests of these methods.

- [x] updates keychain during fetch (Close #37), fetchAll, delete, deleteAll
- [x] add default callbackQueues to async Queries so user doesn't need to specify them
- [x] Bug fix: during a synchronous save of a `ParseObject` if the parse-server returned a ParseError, the framework wouldn't complete on the current thread properly.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -5,10 +5,10 @@ coverage:
   status:
     patch:
       default:
-        target: auto
+        target: 33
     changes: false
     project:
       default:
-        target: 71
+        target: 65
 comment:
   require_changes: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,6 +107,4 @@ jobs:
    steps:
      - uses: actions/checkout@v2
      - name: Carthage 
-       run: carthage build --no-skip-current
-       env: 
-        DEVELOPER_DIR: ${{ env.CI_XCODE_VER }}
+       run: carthage.sh build --no-skip-current
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,4 +107,4 @@ jobs:
    steps:
      - uses: actions/checkout@v2
      - name: Carthage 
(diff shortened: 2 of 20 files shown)


cbaker6:

novemTeam:
Any news?

PR #39, opened 2020-12-22
Change GeoPoint to ParseGeoPoint

cbaker6:
This makes `ParseGeoPoint` consistent with the rest of the SDK.

`File` will be changed to `ParseFile` in a future PR.

### DIFF
--- a/ParseSwift.playground/Pages/7 - GeoPoint.xcplaygroundpage/Contents.swift
+++ b/ParseSwift.playground/Pages/7 - GeoPoint.xcplaygroundpage/Contents.swift
@@ -14,7 +14,7 @@ struct GameScore: ParseObject {
     var createdAt: Date?
     var updatedAt: Date?
     var ACL: ParseACL?
-    var location: GeoPoint?
+    var location: ParseGeoPoint?
     //: Your own properties
     var score: Int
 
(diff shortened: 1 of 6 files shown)


cbaker6:
At one point, I wanted objects created in the SDK to be named `GeoPoint` (without "Parse" in front) and the protocols to `ParseObject` (with "Parse" in front), but this currently isn't possible because of `ACL` in `ParseObject`. The way the `ParseObject` is currently defined, we don't don't need to specify the `CodingKeys` directly since it conforms to the `Codable` protocol, which can encode/decode the property `ACL` directly. If we wanted to rename the property `ACL` to `ParseACL` (allowing us to what I originally wanted with `GeoPoint`), then we would need a custom decoder which would open up a can of worms that isn't worth messing with.

So now, all structs and protocols will have "Parse" in front, which seems to match the JS SDK. 

@TomWFox what do you think?

TomWFox:
I can't lie, I've read this a few times and I can't make sense of it 😅. I think having "Parse" in front is perfectly fine, and not worth opening up a can of worms.

Happy to go with your judgement.

cbaker6:
Lol, I was deep in the code when I wrote this, but this will make everything more consistent.

The only parse type that won't have "Parse" in front after this will be "Query", but a Query is always tied to a "ParseObject", so I don't think will be needed as it will be redundant. Developers won't be able to tie "Querys" to non-parse objects, so it will be okay

PR #40, opened 2020-12-25
Add File support

### LINKED ISSUE #31: Saving ParseObjects with File properties
Right now I am trying to save an ParseObject with a File (ParseFile) property as this:

```swift
    struct TestClass: ParseObject {
        var objectId: String?
        var createdAt: Date?
        var updatedAt: Date?
        var ACL: ParseACL?
        var location: GeoPoint?
    
        var name: String
        var age: Int
        var photo: File
    
        init(name: String, age: Int) {
            self.name = name
            self.age = age
        }
    }
```

It gives me two errors:
Type 'TestClass' does not conform to protocol 'Equatable'
Type 'TestClass' does not conform to protocol 'Hashable'

The first one I can fix by:
`
    static func == (lhs: TestClass, rhs: TestClass) -> Bool {
        return lhs.objectId == rhs.objectId
    }`

How exactly can I save such a property?

cbaker6:
Adds full support for files. Close #31 

- [x] Added test cases
- [x] Saving `ParseFile`'s as child objects
- [x] File upload/download cancellation (works): Some comments on [stack overflow](https://stackoverflow.com/questions/26694651/nsurlsession-delegates-not-called) mention this doesn't work because a completion handlers and delegates can't be used at the same time for `URLSession`, but it's working in this PR.
- [x] Playground examples
- [x] Documentation 
- [x] Switched delete for object batches so they look like regular deletes (removed result enum for ParseError?)
- [x] Placed all server responses in one file Responses.swift
- [x] Bumped the waiting times for async tests from 10 seconds to 20 seconds due to slow CI. Some of the servers for the CI run slow on multithreaded tests and exceed the 10 seconds causing tests to fail. 
- [x] Bug fix: there was an issue of the local ACL being overwritten with `nil` during a save of a new `ParseObject`


novemTeam:
I see the documentation is finished, where can I find it?

cbaker6:
> I see the documentation is finished, where can I find it?

The documentation is here: http://parseplatform.org/Parse-Swift/api/

The docs for “ParseFile” won’t populate until this PR is merged.

cbaker6:
@TomWFox can you look over the added documentation? Particularly in ParseFile.swift. I also added a how-to-use "Files" in playgrounds which can probably read better also.

@pranjalsatija can you review?

My apologies for the big PR, it took a lot to get ParseFile together.

novemTeam:
How can I get the data of a file? In objective c it was like this:

```
PFFileObject *dataFile = queriedFile[projectName];
[dataFile getDataInBackgroundWithBlock:^(NSData * _Nullable data, NSError * _Nullable error) {
}];
```

It looks like it's different with Swift.

cbaker6:
The playgrounds files have examples of how to do some of the general things:

https://github.com/parse-community/Parse-Swift/blob/3f1c897c9b9ddaacaf42e8888131d9059f1d73c0/ParseSwift.playground/Pages/9%20-%20Files.xcplaygroundpage/Contents.swift#L93-L149

PR #42, opened 2020-12-26
Add FieldOperation UnitTests

novemTeam:
Needs to be extended


novemTeam:
Need some help with `mergeWithPrevious`

cbaker6:
That operation doesn't appear to have an equivalent currently in Parse-Swift which is okay. You can look through the comments in https://github.com/parse-community/Parse-SDK-iOS-OSX/blob/5dad4f224d5ae4b36ead891acbdfc15b86d527ae/Parse/Parse/Internal/FieldOperation/PFFieldOperation.h to see if something is named differently or has an equivalent.

For the ones that have comments, you can copy and modify the comments (e.g `PF` to `Parse`) to the `ParseMutationContainer.swift` file so the docs will be populated for the methods.

cbaker6:
A sidenone, be sure to type `git rm -f .DS_Store` in the local repo folders terminal as that file should be removed from the PR before a merge.

novemTeam:
`encoder` must be checked

cbaker6:
Many of these seem like they will be structured similar to the Query tests because they need to ensure that they are encoded to JSON properly. You should look here for an example: https://github.com/parse-community/Parse-Swift/blob/739db445a7850beddc0e63d69c3b8b840f8ca85c/Tests/ParseSwiftTests/ParseQueryTests.swift#L588-L610

novemTeam:
Currently don't have time to continue this, I will come back when this busy period is over.

PR #45, opened 2021-01-04
LiveQuery Support

cbaker6:
Minimum requirements to use LiveQuery are: macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0

Was able to reuse some of the ParseLiveQuery-iOS SDK.

This is a work in progress...

- [x] LiveQuerySocket - Uses `URLSessionWebSocket` hence the minimum requirements above... Uses a dedicated session `URLSession.liveQuery`. Supports multiple connections to different servers
- [x] Delegate authentication challenges to the developer. They can do their own cert pinning and auth if they want. If no authentication provided, will default to OS auth. 
- [x] (Parse) All parse network calls now have a dedicated session, `URLSession.parse` which delegates authentication challenges also. LiveQuery will default to the parse authentication challenge if one isn't specified in the LiveQuery delegate.
- [x] (Parse) Use `POST` instead of `GET` for user login
- [x] LiveQuery Subscribe/Unsubscribe/Update
- [x] LiveQuery Events
- [x] Documentation 
- [x] Playground examples
- [x] Testcases for encoding/decoding, states


cbaker6:
@TomWFox can you review the docs? This PR is moving into testing.

cbaker6:
@pmmlo Can you test this locally?

I have a [server setup](https://github.com/netreconlab/parse-hipaa/tree/parse-swift) via docker that will work out-of-the box with the Playgrounds files if you need it. It also gives you dashboard so you can add/remove/edit fields in there.

The Playgrounds file with how to use LiveQuery is here: https://github.com/parse-community/Parse-Swift/blob/988b0d141df1453fc1c01dc268e49f8f3b2af7bc/ParseSwift.playground/Pages/11%20-%20LiveQuery.xcplaygroundpage/Contents.swift#L33-L88

From what I've seen Subscribe and all of the Events seem to work fine.

I broke something in unsubscribe and update, but it's probably an easy fix I'l get to later.

pmmlo:
> @pmmlo Can you test this locally?

Smooth as butter.  Just changed update query and unsubscribe at the bottom to try-catch but doesn't matter.  Gist here if you care: [https://gist.github.com/pmmlo/a40957e1f51ffaacf47c511cc2a9e034](url)

Add, update, and delete were instantly broadcasted.

Thanks @cbaker6 !

cbaker6:
@pmmlo great to hear! I fixed unsubscribe recently.

pmmlo:
> What do you think about the process of using LiveQuery and subscribing/unsubscribing? I'm hoping it feel more natural...

I'm not the best person to ask, but I think it handles A LOT for the devs and was very easy to use with the Playground.  I personally like it, and it makes sense to me.

Worked very well during my local testing.  It feels more natural than the QueryConstraints framework, but I even learned to like that.  I'll let you know if I get struck by inspiration on improving this while I use it 😏

cbaker6:
@pranjalsatija can you review?

Sources/ParseSwift/LiveQuery/ParseLiveQuery.swift
```diff
@@ -0,0 +1,785 @@
...
+        }
+    }
+
+    //Subscription
+    let requestIdGenerator: () -> RequestId
+    var subscriptions = [RequestId: SubscriptionRecord]()
+    var pendingQueue = [(RequestId, SubscriptionRecord)]() // Behave as FIFO to maintain sending order
+
+    /**
+     - parameter serverURL: The URL of the Parse Live Query Server to connect to.
```
TomWFox, line 151:
```suggestion
     - parameter serverURL: The URL of the `ParseLiveQuery` Server to connect to.
```

Sources/ParseSwift/LiveQuery/ParseLiveQuery.swift
```diff
@@ -0,0 +1,785 @@
...
+        }
+    }
+    #endif
+}
+
+// MARK: Connection
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension ParseLiveQuery {
+
+    /// Manually establish a connection to the `ParseLiveQuery` server.
```
TomWFox, line 486:
```suggestion
    /// Manually establish a connection to the `ParseLiveQuery` Server.
```

Sources/ParseSwift/LiveQuery/ParseLiveQuery.swift
```diff
@@ -0,0 +1,785 @@
...
+                    .asyncAfter(deadline: .now() + DispatchTimeInterval
+                                    .seconds(reconnectInterval)) {
+                    self.createTask()
+                    self.attempts += 1
+                }
+            }
+        }
+    }
+
+    /// Manually disconnect from the `ParseLiveQuery` server.
```
TomWFox, line 519:
```suggestion
    /// Manually disconnect from the `ParseLiveQuery` Server.
```

Sources/ParseSwift/LiveQuery/Protocols/ParseLiveQueryDelegate.swift
```diff
@@ -0,0 +1,69 @@
...
+import Foundation
+
+// swiftlint:disable line_length
+
+///Receive/respond to notifications from the ParseLiveQuery Server.
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+public protocol ParseLiveQueryDelegate: AnyObject {
+
+    /**
+     Respond to authentication requests from a ParseLiveQuery server. If you become a delegate
```
TomWFox, line 18:
```suggestion
     Respond to authentication requests from a ParseLiveQuery Server. If you become a delegate
```

Sources/ParseSwift/LiveQuery/Protocols/SubscriptionHandlable.swift
```diff
@@ -0,0 +1,38 @@
...
+ You can use this protocol on any custom class of yours, instead of Subscription, if it fits your use case better.
+ */
+public protocol  SubscriptionHandlable: AnyObject {
+    /// The type of the `ParseObject` that this handler uses.
+    associatedtype Object: ParseObject
+
+    var query: Query<Object> {get set}
+
+    /**
+     Tells the handler that an event has been received from the `ParseLiveQuery` server.
```
TomWFox, line 22:
```suggestion
     Tells the handler that an event has been received from the `ParseLiveQuery` Server.
```

Sources/ParseSwift/LiveQuery/Subscription.swift
```diff
@@ -0,0 +1,151 @@
...
+//
+//  Created by Corey Baker on 1/2/21.
+//  Copyright © 2021 Parse Community. All rights reserved.
+//
+//
+
+import Foundation
+
+/**
+ Represents an update on a specific object from the `ParseLiveQuery` server.
```
TomWFox, line 13:
```suggestion
 Represents an update on a specific object from the `ParseLiveQuery` Server.
```

Sources/ParseSwift/LiveQuery/Subscription.swift
```diff
@@ -0,0 +1,151 @@
...
+
+    /**
+     Register a callback for when an event occurs of a specific type
+     Example:
+         subscription.handle(Event.Created) { query, object in
+            // Called whenever an object is creaated
+         }
+     - parameter eventType: The event type to handle. You should pass one of the enum cases in `Event`
+     - parameter handler:   The callback to register
+     - returns: The same subscription, for easy chaining
```
TomWFox, line 136:
```suggestion
     - parameter eventType: The event type to handle. You should pass one of the enum cases in `Event`.
     - parameter handler: The callback to register.
     - returns: The same subscription, for easy chaining.
```

Sources/ParseSwift/Storage/PrimitiveObjectStore.swift
```diff
@@ -8,10 +8,22 @@
 import Foundation
 
 // MARK: PrimitiveObjectStore
+/**
+ A store that supports key/value storage. It should be able
+ to handle any object that conforms to encodable and decodable.
+ */
 public protocol PrimitiveObjectStore {
+    /// Delete an object from the store.
+    /// - parameter key: The unique key value of the object
```
TomWFox, line 17:
```suggestion
    /// - parameter key: The unique key value of the object.
```

Sources/ParseSwift/Types/Query.swift
```diff
@@ -664,6 +665,38 @@ public class Query<T>: Encodable, Equatable where T: ParseObject {
         return self
     }
 
+    /**
+     A variadic list of fields to receive when receiving a `ParseLiveQuery`.
+     
+     Suppose the `ParseObject` Player contains three fields name, id and age.
+     If you are only interested in the change of the name field, you can set query.fields to "name".
```
TomWFox, line 672:
```suggestion
     If you are only interested in the change of the name field, you can set `query.fields` to "name".
```

Sources/ParseSwift/Types/Query.swift
```diff
@@ -664,6 +665,38 @@ public class Query<T>: Encodable, Equatable where T: ParseObject {
...
+    public func fields(_ keys: String...) -> Query<T> {
+        self.fields = keys
+        return self
+    }
+
+    /**
+     A list of fields to receive when receiving a `ParseLiveQuery`.
+     
+     Suppose the `ParseObject` Player contains three fields name, id and age.
+     If you are only interested in the change of the name field, you can set query.fields to "name".
```
TomWFox, line 688:
```suggestion
     If you are only interested in the change of the name field, you can set `query.fields` to "name".
```

TomWFox (changes requested):
Only being picky really...

pmmlo:
@cbaker6 I had a little time to test this with a HTTPS server and `.subscribe` and `.handleSubscribe` return expected callbacks.  However, after about ~29s (timed twice, but seems consistent), I get these two errors:
```
2021-01-15 16:12:45.010269-0800 Member[60224:14126091] [tcp] tcp_input [C1.1:3] flags=[R] seq=3454113186, ack=0, win=0 state=LAST_ACK rcv_nxt=3454113186, snd_una=1642147939
2021-01-15 16:12:45.011230-0800 Member[60224:14126091] [tcp] tcp_input [C1.1:3] flags=[R] seq=3454113186, ack=0, win=0 state=CLOSED rcv_nxt=3454113186, snd_una=1642147939
2021-01-15 16:12:45.012326-0800 Member[60224:14126091] [tcp] tcp_input [C1.1:3] flags=[R] seq=3454113186, ack=0, win=0 state=CLOSED rcv_nxt=3454113186, snd_una=1642147939
```

It seems like the Subscription is timing out or crashing.  I have guesses on what may be going on but thought you may know immediately.

Here's the code:

```
query = Team.query("member" == User.current?.objectId)
guard let subscription = query.limit(1).subscribe else { return }

subscription.handleSubscribe { subscribedQuery, isNew in
    if isNew {
        print("Successfully subscribed to new query \(subscribedQuery)")
    } else {
        print("Successfully updated subscription to new query \(subscribedQuery)")
    }
}

subscription.handleEvent { _, event in
    print("\(event)")
    switch event {
    case .entered(let object):
        print("Entered: \(object)")
    case .left(let object):
        print("Left: \(object)")
    case .created(let object):
        print("Created: \(object)")
    case .updated(let object):
        print("Updated: \(object)")
    case .deleted(let object):
        print("Deleted: \(object)")
    }
}

 subscription.handleUnsubscribe { query in
    print("Unsubscribed from \(query)")
}
```

I never receive any events, despite editing in dashboard.  I can unsubscribe with the expected response before the two errors fire.

Any ideas?

cbaker6:
Did you register your `Team` class for LiveQuery in your index file of your Parse Server?

```
liveQuery: {
    classNames: ["GameScore", "Book", "Author"] // List of classes to support for query subscriptions
  },
```

I tested this LiveQuery by leaving a subscriptions on over night in the simulator for an iPhone and Apple Watch and it didn't disconnect and sent events the next day.

cbaker6:
You can see how I setup the server for playgrounds here: https://github.com/netreconlab/parse-hipaa/blob/parse-swift/parse/index.js

pmmlo:
> You can see how I setup the server for playgrounds here: https://github.com/netreconlab/parse-hipaa/blob/parse-swift/parse/index.js

I have something similar with Express as well.  I also ensured livequery server was initialized and the `Team` class name was listed.

I guess not as straight forward as I thought it would be.  I will dig into this then.  Must be an implementation issue.

pmmlo:
> Also, for your server, are you behind a proxy? If you are using Nginx be sure to setup https://www.nginx.com/blog/websocket-nginx/

I'm actually running it in a k8s cluster with an ingress.

cbaker6:
But your http works when you use playgrounds and docker?

pmmlo:
Yup, even graphql.

pmmlo:
I ran playgrounds using your docker and playgrounds.  Worked well there.

cbaker6:
Maybe it's configuration, can you check if https://github.com/parse-community/Parse-Swift/blob/84f923141b0994dc50bb924cc8fbcb6fad228158/Sources/ParseSwift/LiveQuery/ParseLiveQuery.swift#L192

is converting your url to `wss` since you are tls?

pmmlo:
> Maybe it's configuration, can you check if
> 
> https://github.com/parse-community/Parse-Swift/blob/84f923141b0994dc50bb924cc8fbcb6fad228158/Sources/ParseSwift/LiveQuery/ParseLiveQuery.swift#L192
> 
> is converting your url to `wss` since you are tls?

This looks promising.  I think it may be...let me try to test this.

pmmlo:
It's wss.

cbaker6:
Are you able to connect using a site like https://www.websocket.org/echo.html

pmmlo:
There was an undocumented backend timeout of 30 seconds in AWS that seems to have been the issue.  Would have guessed 30 seconds was enough...  Just raised it to 5 minutes, which should be more than enough.  No error, but I'm still not getting anything from .handleEvent(). I'll play around with this and post, if I figure out my issue. ;)

pmmlo:
Just started working after I put in breakpoints.  Changed nothing, and I have an empty console.  Maybe I should have cleaned the project first...

Sorry!

PR #48, opened 2021-01-08
Bug fix: issue when encoding objects that are already saved

### LINKED ISSUE #47: New issue with .saveAll batch operation
Sorry @cbaker6 but I'm having a new issue with .saveAll batch save/update operation since one of the last two commits.

I can't imagine it's because of the cloudcode name changes from #46 but maybe there is a cloud code that runs in parse-server source that I am not aware of.  I scanned through #43 and I'm going to take a flyer and say it might be the skippedKeys or other changes to the `batch` command in API+Commands.swift.  I don't see the error but I don't know enough about the return headers and body to know how it is encoded.

Also, I think Parse has always done this, but I get a success response with an array of parseerrors.  So, it's a success but the saveAll failed.

Here's the code:

```swift
/// I have something like this where GameScore.scores is a [String]?
let score = GameScore(score: ["2"])
let score2 = GameScore(score: ["3"])
let scoresArray = [score, score2]

scoresArray.saveAll { results in
    switch results {
    case .success(let success):
        var index = 0
        success.forEach { success in
            switch success {
            case .success(let savedScore):
                print("Saved \"\(savedScore.className)\" with score \(savedScore.score) successfully")
                if index == 1 {
                    print("\(savedScore)")
                }
                index += 1
            case .failure(let error):
                print("Error saving: \(error)") // <---All the objects fail now after the recent commits and scores are not updated remotely.
            }
        }

    case .failure(let error):
        print("Error batch saving: \(error)")
    }
}
```

```swift
// This is the response from the `error` in the first try-catch
[Swift.Result<ParseApp.GameScore, ParseSwift.ParseError>.failure(ParseSwift.ParseError(code: ParseSwift.ParseError.Code.invalidKeyName, message: "Invalid field name: __type.")), Swift.Result<ParseApp.GameScore, ParseSwift.ParseError>.failure(ParseSwift.ParseError(code: ParseSwift.ParseError.Code.invalidKeyName, message: "Invalid field name: __type."))]
```

I thought there may be some issue with a pointer, missing column, or my key strings in query, because `Invalid field name:__type.` in the response, but after some inspection it seems to be a potential issue with encoding the query.  Everything is in order and looking at history, it seemed to work before.  I could be wrong of course.  I may try rolling back and testing, but for now, it would be nice to have another pair of eyes who knows Parse-Swift well.

There are no issues updating remote objects, if I do a for-loop and do individual `.save`.

cbaker6:
Fixes #47. Previous commits didn't enforce `ParseObjects` going through the ParseEncoder. Since they go through now, there was a bug in the encoder when objects were already saved. They were automatically converted to Pointers, even if they were themselves. This caused issues when attempting to saveAll as 1) there were keys the server didn't like 2) No other key/value info wasn't being sent during the update

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -5,7 +5,7 @@ coverage:
   status:
     patch:
       default:
-        target: 69
+        target: auto
     changes: false
     project:
       default:
--- a/Sources/ParseSwift/Coding/ParseEncoder.swift
+++ b/Sources/ParseSwift/Coding/ParseEncoder.swift
@@ -262,7 +262,9 @@ private class _ParseEncoder: JSONEncoder, Encoder {
                 throw ParseError(code: .unknownError, message: "Found a circular dependency when encoding.")
             }
             self.uniqueObjects.insert(object)
-            if !self.collectChildren {
+            if !self.collectChildren && codingPath.count > 0 {
+                valueToEncode = value
+            } else if !self.collectChildren {
                 valueToEncode = value.toPointer()
             }
         } else {
--- a/Sources/ParseSwift/LiveQuery/ParseLiveQuery.swift
+++ b/Sources/ParseSwift/LiveQuery/ParseLiveQuery.swift
@@ -683,7 +683,7 @@ extension ParseLiveQuery {
     }
 }
 
-// MARK: Query - Subscribe
+// MARK: ParseLiveQuery - Subscribe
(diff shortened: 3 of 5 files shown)


cbaker6:
@pmmlo thanks for filing this issue. It was definitely a corner case that wasn't easy to find.

Can you test this out in your scenarios and left me know if you see any other issues. You should try to run other scenarios that you know use to work as well.

PR #53, opened 2021-01-15
Anonymous and Apple Login

cbaker6:
This PR adds 3rd party login ability via the `ParseAuthenticatable` protocol.

- [x] Add `ParseAnonymous` and `ParseApple` login
- [x] Add `ParseAuthenticatable` protocol for implementation by developers. Example: Facebook, Twitter, etc. should conform to protocol and PR's can be submitted to add them. The dependency of the the respective framework, i.e. `import FBSDKCoreKit` shouldn't be part of the PR (this should be done by the developer in their respective apps). Instead the implementation should just ensure the necessary `AuthenticationKeys` are captured to properly authenticate the respective framework with a Parse Server. An example of this is [ParseApple.swift](https://github.com/parse-community/Parse-Swift/blob/login/Sources/ParseSwift/Authentication/3rd%20Party/ParseApple.swift). As a starting point for those who want to add more 3rd party methods, I recommend copying `ParseApple.swift` and modifying `AuthenticationKeys` and helper methods. All of the Parse server-side 3rd party supported authentication methods are [here](https://github.com/parse-community/parse-server/tree/033a0bd443ba846fa5a7fae969eb8f71cf492162/src/Adapters/Auth) along with their respective [documentation](https://docs.parseplatform.org/parse-server/guide/#oauth-and-3rd-party-authentication). 
- [x] Add `ParseSession` protocol 
- [x] Add documentation
- [x] Add test cases
- [x] Add playground examples    

Usage (same can be done for "apple", just need to pass in apple credentials to login):

```swift
// Anonymous asynchronous login

//Assuming you define the ParseUser
struct User: ParseUser {
    //: These are required for ParseObject
    var objectId: String?
    var createdAt: Date?
    var updatedAt: Date?
    var ACL: ParseACL?

    //: These are required for ParseUser
    var username: String?
    var email: String?
    var password: String?

    //: Your custom keys
    var customKey: String?
}

// You can log the User in using:
User.anonymous.login { result in
    switch result {
    case .success:
        print("Successfully logged in \(User.current)")
    case .failure(let error):
        print("Error logging in: \(error)")
    }
}

### CODE REVIEW

**Comment by cbaker6** [comment]
@TomWFox can you look over the docs when you get a chance? I'm pretty much done with them and I don't think I need to a add anymore methods to the login protocols

**Comment by cbaker6** [comment]
@TomWFox we should probably add an improved version of my second bullet point above to the contributors guide (and maybe a snippet to the readme?):

> Add `ParseAuthenticatable` protocol for implementation by developers. Example: Facebook, Twitter, etc. should conform to protocol and PR's can be submitted to add them. The dependency of the the respective framework, i.e. `import FBSDKCoreKit` shouldn't be part of the PR (this should be done by the developer in their respective apps). Instead the implementation should just ensure the necessary `AuthenticationKeys` are captured to properly authenticate the respective framework with a Parse Server. An example of this is [ParseApple.swift](https://github.com/parse-community/Parse-Swift/blob/login/Sources/ParseSwift/Authentication/3rd%20Party/ParseApple.swift). As a starting point for those who want to add more 3rd party methods, I recommend copying `ParseApple.swift` and modifying `AuthenticationKeys` and helper methods. All of the Parse server-side 3rd party supported authentication methods are [here](https://github.com/parse-community/parse-server/tree/033a0bd443ba846fa5a7fae969eb8f71cf492162/src/Adapters/Auth) along with their respective [documentation](https://docs.parseplatform.org/parse-server/guide/#oauth-and-3rd-party-authentication). 

I can see a few questions/issues occurring frequently:

1. Can support be added for a "specific" authentication type? (Yes, following the directions above, developers are encouraged to submit PR's and have a blueprint on how to make a new one).
2. This is different from the objc-c SDK, can we get PFFacebookUtils and PFFacebookUtils? (Not exactly, the dependency won't be added, but developers are encouraged to create and conform to `ParseAuthenticatable`

I think the method above we have setup for adding support for basically any authentication the server supports leads to easier maintainability. Plus, this allows users the flexibility of using any version of their authentication framework they want, they simply just need to provide the minimum keys.

File: Sources/ParseSwift/Authentication/3rd Party/ParseApple.swift [hunk 1]
```diff
@@ -0,0 +1,150 @@
...
+
+    /// Authentication keys required for Apple authentication.
+    enum AuthenticationKeys: String, Codable {
+        case id // swiftlint:disable:this identifier_name
+        case token
+
+        /// Properly makes an authData dictionary with the required keys.
+        /// - parameter id: Required id.
+        /// - parameter token: Required token.
+        /// - returns: required authData dictionary.
```
TomWFox, line 28:
```suggestion
        /// - returns: Required authData dictionary.
```

Sources/ParseSwift/Objects/ParseSession.swift
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     The session token for this session.
+     */
+    var sessionToken: String { get }
+
+    /// The user the session is for.
+    var user: SessionUser { get }
+
+    /// Whether the session is restricted.
+    var restricted: Bool { get }
```
TomWFox, line 28: I think this could be an issue, see https://github.com/parse-community/parse-server/issues/6612

Sources/ParseSwift/Authentication/Protocols/ParseAuthenticatable.swift
```diff
@@ -0,0 +1,323 @@
...
+    /**
+     Strips the *current* user of a respective authentication type.
+     - returns: the *current* user whose autentication type was stripped. Returns `nil`
+     if there's no current user. This modified user has not been saved.
+     */
+    func strip()
+
+    /**
+     Strips the `ParseUser`of a respective authentication type.
+     - parameter user: the `ParseUser` to strip. The user must be logged in on this device.
```
TomWFox, line 86:
```suggestion
     - parameter user: The `ParseUser` to strip. The user must be logged in on this device.
```

Sources/ParseSwift/Authentication/Protocols/ParseAuthenticatable.swift
```diff
@@ -0,0 +1,323 @@
...
+     - returns: the user whose autentication type was stripped. This modified user has not been saved.
```
TomWFox, line 87:
```suggestion
     - returns: The user whose autentication type was stripped. This modified user has not been saved.
```

Sources/ParseSwift/Authentication/Protocols/ParseAuthenticatable.swift
```diff
@@ -0,0 +1,323 @@
...
+     - parameter completion: The block to execute.
+     It should have the following argument signature: `(Result<Self, ParseError>)`.
+     */
+    func unlink(options: API.Options,
+                callbackQueue: DispatchQueue,
+                completion: @escaping (Result<AuthenticatedUser, ParseError>) -> Void)
+
+    /**
+     Strips the *current* user of a respective authentication type.
+     - returns: the *current* user whose autentication type was stripped. Returns `nil`
```
TomWFox, line 79:
```suggestion
     - returns: The *current* user whose autentication type was stripped. Returns `nil`
```

Sources/ParseSwift/Authentication/Protocols/ParseAuthenticatable.swift
```diff
@@ -0,0 +1,323 @@
...
+     Whether the `ParseUser` is logged in with the respective authentication type.
+     - parameter user: the `ParseUser` to check authentication type. The user must be logged in on this device.
+     - returns: `true` if the `ParseUser` is logged in via the repective
+     authentication type. `false` if the user is not.
+     */
+    func isLinked(with user: AuthenticatedUser) -> Bool
+
+    /**
+     Unlink the `ParseUser` *asynchronously* from the respective authentication type.
+     - parameter user: the `ParseUser` to unlink. The user must be logged in on this device.
```
TomWFox, line 55:
```suggestion
     - parameter user: The `ParseUser` to unlink. The user must be logged in on this device.
```

Sources/ParseSwift/Authentication/Protocols/ParseAuthenticatable.swift
```diff
@@ -0,0 +1,323 @@
...
+     - parameter completion: The block to execute.
+     */
+    func link(authData: [String: String]?,
+              options: API.Options,
+              callbackQueue: DispatchQueue,
+              completion: @escaping (Result<AuthenticatedUser, ParseError>) -> Void)
+
+    /**
+     Whether the `ParseUser` is logged in with the respective authentication type.
+     - parameter user: the `ParseUser` to check authentication type. The user must be logged in on this device.
```
TomWFox, line 47:
```suggestion
     - parameter user: The `ParseUser` to check authentication type. The user must be logged in on this device.
```

Sources/ParseSwift/Authentication/Internal/ParseAnonymous.swift
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// MARK: Login
+public extension ParseAnonymous {
+    /**
+     Login a `ParseUser` *synchronously* using the respective authentication type.
+     - parameter authData: The authData for the respective authentication type.
+     - parameter options: A set of header options sent to the server. Defaults to an empty set.
+     - throws: `ParseError`.
+     - returns the linked `ParseUser`.
```
TomWFox, line 51:
```suggestion
     - returns: the linked `ParseUser`.
```

TomWFox (changes requested):
Mostly minor changes with one potential issue.

Getting closer to 1.0!

Sources/ParseSwift/Objects/ParseSession.swift
cbaker6: I see, I think making it "optional" should make it backwards/forwards compatible with Parse Server. What do you think?

cbaker6: I can also remove it, developers just won't be able to see it in their Swift apps because it's never decoded

cbaker6: The only potential problem I see with removing is it will show in their Parse Dashboard, but not on their swift apps...

cbaker6:
Guessing at the issue with jazzy docs.

I believe GitHub Actions may have recently bumped the default Xcode to 12.3 and I "think" jazzy and it's dependencies aren't playing nice. It looks like the docs are building fine with 12.2, so docs can stay on that version for now, while the rest of the builds continue on the latest Xcode.

TomWFox:
> @TomWFox we should probably add an improved version of my second bullet point above to the contributors guide (and maybe a snippet to the readme?):

I think adding to the contributors guide would be a good idea, not sure about the readme.

Sources/ParseSwift/Objects/ParseSession.swift
TomWFox: Making it optional seems like an acceptable option. I wonder whether the comment should be changes to say that the feature isn't implemented server side and/or link to the #6612

PR #54, opened 2021-01-17
Add Roles and Relations

cbaker6:
- [x] Finished `ParseOperation` along with helper properties/methods to `ParseObjects`.
- [x] Create  `ParseRole`
- [x] Query - added `distinct`, `aggregate` (aggregate needs testing on the server), `nor`, `containedBy`, option not to sort near for GeoQueries, relative time query. Switched `Query` to value type. See below...
- [x] Create `ParseRelation` along with helper properties/methods to `ParseObjects`. 
- [x] Documentation
- [x] Improve project naming conventions and structure 
- [x] Add test cases for added functionality
- [x] Fixed bugs in batch object test cases related to comparing dynamic dates incorrectly that caused random test failures.
- [x] Playground examples 

*Possible* breaking changes:
- `Query` has been switched to a value type for improved thread safety. You are still able to chain queries, but you may have some warnings to switch from `let` to `var`. In addition, if you depended on the reference type of `Query`, you may need to set the return value to the old value
- `Pointer`and `toPointer` - removed fatal error crash when missing `objectId`. This `throws`  an error instead. The compiler may request for you to add `try`'s in your code depending on your implementation.
- `ParseUser` - the `authData` property was added to `ParseUser` for authentication in the previous PR. The compiler should force any object that conforms to `ParseUser` to add the property.


cbaker6:
@TomWFox can you look over the docs?

I'm working on the rest of the tests for Role and Relation, but I believe I have most if not all the docs in

Sources/ParseSwift/Types/Query.swift
```diff
@@ -593,46 +597,51 @@ public class Query<T>: Encodable, Equatable where T: ParseObject {
...
-        self.include = ["*"]
-        return self
+        var mutableQuery = self
+        mutableQuery.include = ["*"]
+        return mutableQuery
     }
 
     /**
-      Executes a distinct query and returns unique values. Default is to nil.
+     Exclude specific keys for a`ParseObject`s. Default is to nil.
```
TomWFox, line 638:
```suggestion
     Exclude specific keys for a `ParseObject`. Default is to nil.
```

Sources/ParseSwift/Types/ParseRelation.swift
```diff
@@ -0,0 +1,224 @@
...
+            throw ParseError(code: .unknownError, message: "All objects have to have the same className.")
+        }
+        return try parent.operation.removeRelation(key, objects: objects)
+    }
+
+    /**
+     Returns a `Query` that is limited to objects in this relation.
+        - parameter child: The child class for the relation.
+        - throws: An error of type `ParseError`.
+        - returns: A relation query
```
TomWFox, line 106:
```suggestion
        - returns: A relation query.
```

Sources/ParseSwift/Operations/ParseOperation.swift
```diff
@@ -0,0 +1,412 @@
...
+
+import Foundation
+
+/**
+ A `ParseOperation` represents a modification to a value in a `ParseObject`.
+ For example, setting, deleting, or incrementing a value are all `ParseOperation`'s.
+ `ParseOperation` themselves can be considered to be immutable.
+ 
+ In most cases, you should not call this class directly as a `ParseOperation` can be
+ indirectly created from any `ParseObject` by using its' `operation` property.
```
TomWFox, line 17:
```suggestion
 indirectly created from any `ParseObject` by using its `operation` property.
```

TomWFox (changes requested):
Only minor requests...

cbaker6:
Thanks @TomWFox! I recently added some of the missing query constraints. Not too many, but can look at them? https://github.com/parse-community/Parse-Swift/pull/54/commits/6bb02bb2f746b4f5164c323812a49553eec0b4e6

Sources/ParseSwift/Types/Query.swift
```diff
@@ -232,7 +245,31 @@ public func notContainedIn <T>(key: String, array: [T]) -> QueryConstraint where
...
+  - parameter key: The key to be constrained.
+  - parameter array: The possible values for the key's object.
+  - returns: The same instance of `QueryConstraint` as the receiver.
+ */
+public func containedBy <T>(key: String, array: [T]) -> QueryConstraint where T: Encodable {
+    QueryConstraint(key: key, value: array, comparator: .containedBy)
+}
+
+/**
+ Add a constraint to the query that requires a particular key's time is related a specified time. E.g. "3 days ago".
```
TomWFox, line 263:
```suggestion
 Add a constraint to the query that requires a particular key's time is related to a specified time. E.g. "3 days ago".
```

TomWFox (changes requested):

PR #56, opened 2021-01-21
Release 1.0.0

cbaker6:
@TomWFox I took the initial shot at the change log. Feel free to modify/improve as you see fit.

Did you get a chance to add your credentials for Cocoapods?

Let me know if I need to add more to prep for the release...

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,9 +27,9 @@ jobs:
         uses: actions/cache@v2
         with:
           path: vendor/bundle
-          key: ${{ runner.os }}-gem-${{ hashFiles('**/Gemfile.lock') }}
+          key: ${{ runner.os }}-gem-v1-${{ hashFiles('**/Gemfile.lock') }}
           restore-keys: |
-            ${{ runner.os }}-gem-
+            ${{ runner.os }}-gem-v1
       - name: Install Bundle
         run: |
           bundle config path vendor/bundle
--- a/ParseSwift.podspec
+++ b/ParseSwift.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = "ParseSwift"
-  s.version  = "0.0.1"
+  s.version  = "1.0.0"
   s.summary  = "Parse Pure Swift SDK"
   s.homepage = "https://github.com/parse-community/Parse-Swift"
   s.author = {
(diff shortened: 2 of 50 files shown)


TomWFox:
I just removed back-end type entries (testing, GitHub etc) - I like to leave these out as they generally aren't relevant to the developer using the SDK.

.github/workflows/release.yml
```diff
@@ -27,9 +27,9 @@ jobs:
         uses: actions/cache@v2
         with:
           path: vendor/bundle
-          key: ${{ runner.os }}-gem-${{ hashFiles('**/Gemfile.lock') }}
+          key: ${{ runner.os }}-gem-v1-${{ hashFiles('**/Gemfile.lock') }}
           restore-keys: |
-            ${{ runner.os }}-gem-
+            ${{ runner.os }}-gem-v1
       - name: Install Bundle
         run: |
```
TomWFox, line 34: Did you mean to add in the documentation deployment in the release workflow?

ParseSwift.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = "ParseSwift"
-  s.version  = "0.0.1"
+  s.version  = "1.0.0"
   s.summary  = "Parse Pure Swift SDK"
   s.homepage = "https://github.com/parse-community/Parse-Swift"
   s.author = {
```
TomWFox, line 6: You could add your name here

cbaker6:
@TomWFox I added ParseConfig, can you look through the docs?

.github/workflows/release.yml
cbaker6: I think I know what you mean... the release workflow is already setup to deploy docs, I just needed to update the cache

cbaker6:
@TomWFox Also, are you able to change "experimental" on the top right of the GitHub page? It won't let me change it.

The only features I can think of that Parse-Swift doesn't have compared to Obj-C and JS are cache and local data store. Parse-Swift does cache and store User, Installation, and Config to the Keychain automatically, but not the rest of the ParseObjects.

When it comes to what Parse-Swift does have, it has more query ability than Obj-c, and seems to be around the same as JS (Parse-Swift does have relativeTime query, which I didn't see in JS, but there may be something in JS I missed, so we can go for similar for now).

Let me know if you think of any features...

cbaker6:
I forgot to mention an important feature. Parse-Swift enables certificate pinning and authentication for both LiveQuery and the standard SDK. It seems like this was a feature request for some years in the Objc-SDK. Not sure if/how the JS SDK handles that

.github/workflows/release.yml
TomWFox: My bad, I saw you removed the docs deploy from the other workflow, didn't see that it was already here!

PR #60, opened 2021-01-23
Release 1.0.2

cbaker6:


TomWFox (approved):

cbaker6:
@TomWFox looks like we gained the Linux badge (finally): https://swiftpackageindex.com/parse-community/Parse-Swift

Looking at the [builds](https://swiftpackageindex.com/parse-community/Parse-Swift/builds), Parse-Swift is compatible with Swift 5.3+. It looks like Swift Linux < Swift 5.3 is missing some important items the SDK needs for encoding. This is a win IMO, as the Swift Linux devs are constantly adding more, so LiveQuery should be available in the near future.

Lastly, I don't think I mentioned, but those same builds show Parse-Swift builds on Apple Silicon as well, so the SDK is future proof for the time being...

PR #65, opened 2021-01-24
Use ParseLiveQuery Subscription as a SwiftUI view model

cbaker6:
Adds support to use a `ParseLiveQuery` subscription as a VM. Once more people start using the SDK, this will probably be a highly sought after feature.

`ParseLiveQuery` and its ability to subscribe to events is suited to be a great candidate for SwiftUI updates or serve as a publisher of data in general to allow updates to views. The original `Subscription`, now renamed to `SubscriptionCallback` was a spin on the original `Subscription` in `ParseLiveQuery-iOS`. The new `Subscription` and the original essentially behave in a similar manner except the original was designed before property wrappers and before SwiftUI.

The change allows devs a different way from the objective-c SDK when accessing Parse data.

1. Use a traditional query find (pull) to get the original objects to display in your SwiftUI view.
2. Use a `ParseLiveQuery` Subscription as a view model (VM) for any subsequent updates in your SwiftUI view.

This update is "slightly" breaking as the original `Subscription` has been renamed to `SubscriptionCallback`. Both still have properties accessible directly from any `Query`. For example if you have `var query = GameScore.query("score" > 9)`. Then:

1. `Subscription` - ` let subscription = query.subscribe`
2. `SubscriptionCallback` - ` let subscription = query.subscribeCallback`


cbaker6:
@TomWFox I added docs to the "new" `Subscription`. Can you review them?

Sources/ParseSwift/LiveQuery/ParseLiveQuery.swift
```diff
@@ -729,6 +739,24 @@ public extension Query {
...
+     */
+    var subscribeCallback: SubscriptionCallback<ResultType>? {
+        try? ParseLiveQuery.client?.subscribe(self)
+    }
+
+    /**
+     Registers the query for live updates, using the default subscription handler,
+     and a specific `ParseLiveQuery` client.
+     - parameter client: A specific client.
+     - returns: The subscription that has just been registered
```
TomWFox, line 755:
```suggestion
     - returns: The subscription that has just been registered.
```

TomWFox (approved):
Just one `.` 😅

LGTM 🚀

PR #73, opened 2021-01-30
Add combine publishers for all objects and types

cbaker6:
Adding the publishers such as `Future` to all one-shot async calls, making Parse-Swift modernized with SwiftUI. All methods that are capable of “publishing” end with the word `Publisher`.  This means that there are 3 ways to use most of the methods that send/receive info to a Parse Server: 1) sync, 2) async, and 3) publisher.

As far as usage, you use the added publishers just like any other `Combine` publisher. The thing to remember is that all of the push/pull methods like `save`, `fetch`, etc. are `Futures`, so they start their work as soon as you call them and will only publish one result when finished. Any new subscriptions to an already fulfilled`Future` will produce the original result. This is different from the `ParseLiveQuery` `Subscription` which will keep publishing events/subscriptions as long as it’s not cancelled.

`Combine` currently isn't supported on Linux, but if/when it ever is the combine features will automatically be available for use. Linux devs can use the async/sync callbacks instead.

- [x] All objects publishable
- [x] All types publishable
- [x] Testcases
- [x] Some fixes to `ParseUser.logout` to make it more predictable 
- [x] Bring minimum deployment to the same level on Xcode, SPM, and Cocoapods, `[.iOS(.v12), .macOS(.v10_13), .tvOS(.v12), .watchOS(.v5)]`. If a developer wants to support an older platform, they should use the [Parse-Obj-C SDK](https://github.com/parse-community/Parse-SDK-iOS-OSX)  

Slightly breaking (compiler should suggest the correct fix):
- `delete`, `deleteAll`, `passwordRest`, and `verificationEmail` use to all return an `ParseError`. These have all been switched to return a Result<Void, ParseError> instead to match everything else. This is helpful, because `Future` expects a result enum anyways


cbaker6:
@TomWFox can you review the docs? All of the added docs are SwiftUI docs.

The large number of line additions is mostly test code

Sources/ParseSwift/Types/ParseFile+combine.swift
```diff
@@ -0,0 +1,94 @@
...
+                       progress: ((URLSessionTask, Int64, Int64, Int64) -> Void)? = nil) -> Future<Self, ParseError> {
+        Future { promise in
+            self.save(options: options,
+                      progress: progress,
+                      completion: promise)
+        }
+    }
+
+    /**
+     Deletes the file from the Parse cloud. Publishes when complete.
```
TomWFox, line 82:
```suggestion
     Deletes the file from the Parse Server. Publishes when complete.
```

TomWFox (approved):
Looks good!

PR #78, opened 2021-02-14
Support for PushNotifications

cbaker6:
- [x] Add `setDeviceToken` to `ParseInstallation` to properly convert `deviceToken` to a hex `String` to be used by Parse Server
- [x] Documentation 
- [x] Testcases
- [x] Changelog entry
- [x] Prepare for next version release (bump project, pod spec, and jazzy) 
- [x] `if #available(macOS 11.0, iOS...)` doesn't work well when an older version OS is looking for a newer version than what it currently knows. This causes the SPM 11.7 (Swift 5.2) test failure. Added compiler check to work across older versions of Swift.

### DIFF
--- a/Sources/ParseSwift/Objects/ParseInstallation.swift
+++ b/Sources/ParseSwift/Objects/ParseInstallation.swift
@@ -95,6 +95,17 @@ extension ParseInstallation {
 
         return .installations
     }
+
+    /**
+     Sets the device token string property from an `Data`-encoded token.
+     - parameter data: A token that identifies the device.
+     */
+    mutating public func setDeviceToken(_ data: Data) {
+        let deviceTokenString = data.hexEncodedString()
+        if deviceToken != deviceTokenString {
+            deviceToken = deviceTokenString
+        }
+    }
 }
 
 // MARK: CurrentInstallationContainer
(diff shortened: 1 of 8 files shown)


cbaker6:
@TomWFox can you look over the docs I added?

 I still need look for the possible fix I mentioned above...

Sources/ParseSwift/Objects/ParseInstallation.swift
```diff
@@ -29,7 +29,8 @@ import AppKit
  latest device state if the installation has not recently been saved.
 
  `ParseInstallation` installations which have a valid `deviceToken` and are saved to
```
TomWFox, line 31:
```suggestion
 `ParseInstallation`s which have a valid `deviceToken` and are saved to
```

Sources/ParseSwift/Objects/ParseInstallation.swift
```diff
@@ -29,7 +29,8 @@ import AppKit
...
- the Parse cloud can be used to target push notifications.
+ the Parse cloud can be used to target push notifications. Use `setDeviceToken` to set the
```
TomWFox, line 32:
```suggestion
 the Parse Server can be used to target push notifications. Use `setDeviceToken` to set the
```

TomWFox (commented):
Nice 👍

PR #88, opened 2021-03-02
Fix query select and exclude

### LINKED ISSUE #86: .select() in finds and firsts does not seem to exclude any fields
Hey @cbaker6  - Do you think this could be an encoding problem or a potential parse-server bug?

```
    var objectId: String?
    var createdAt: Date?
    var updatedAt: Date?
    var ACL: ParseACL?

    var score: Int?
}

var score = GameScore()
score.score = 200
try score.save()

let afterDate = Date().addingTimeInterval(-300)
var query = GameScore.query("createdAt" > afterDate).select(["score"]).limit(1)

query.find() { results in
    print(results) /// <<< I would expect only the GameScore objects with createdAt after afterDate to only return the "score" field. (Yes, this is lazy.)
}

/// Alternatively
query.first() { results in
    print(results)
}
```

Anybody have any thoughts?  Will dig deeper into the source, if not.  Thanks!

EDIT:
.select() does seem to work, as of late.  Reference: [https://stackoverflow.com/questions/61100282/parse-server-select-a-few-fields-from-included-object](https://stackoverflow.com/questions/61100282/parse-server-select-a-few-fields-from-included-object)

cbaker6:
Fix #86 select and exclude https://github.com/parse-community/parse-server/pull/7242 

- [x] Add variadic version of `exclude`
- [x] `select` and `exclude` example in playground
- [x] Ensure thread-safe `endpoint` is used for queries
- [x] Modify fetch include to use an encoded array instead of join array   
- [x] Improve test cases `select`, `exclude`, and fetch include

### DIFF
--- a/Sources/ParseSwift/Types/Query.swift
+++ b/Sources/ParseSwift/Types/Query.swift
@@ -546,14 +546,14 @@ public struct Query<T>: Encodable, Equatable where T: ParseObject {
     private let method: String = "GET"
     internal var limit: Int = 100
     internal var skip: Int = 0
-    internal var keys: String? //[String]?
+    internal var keys: [String]?
     internal var include: [String]?
     internal var order: [Order]?
     internal var isCount: Bool?
     internal var explain: Bool?
     internal var hint: String?
     internal var `where` = QueryWhere()
-    internal var excludeKeys: String? //[String]?
+    internal var excludeKeys: [String]?
     internal var readPreference: String?
     internal var includeReadPreference: String?
     internal var subqueryReadPreference: String?
@@ -701,7 +701,7 @@ public struct Query<T>: Encodable, Equatable where T: ParseObject {
      */
     public func exclude(_ keys: String...) -> Query<T> {
         var mutableQuery = self
-        mutableQuery.excludeKeys = keys.joined(separator: ",")
+        mutableQuery.excludeKeys = keys
         return mutableQuery
     }
 
@@ -711,7 +711,7 @@ public struct Query<T>: Encodable, Equatable where T: ParseObject {
     */
     public func exclude(_ keys: [String]) -> Query<T> {
         var mutableQuery = self
-        mutableQuery.excludeKeys = keys.joined(separator: ",")
+        mutableQuery.excludeKeys = keys
         return mutableQuery
     }
 
@@ -722,7 +722,7 @@ public struct Query<T>: Encodable, Equatable where T: ParseObject {
      */
     public func select(_ keys: String...) -> Query<T> {
         var mutableQuery = self
-        mutableQuery.keys = keys.joined(separator: ",")
+        mutableQuery.keys = keys
         return mutableQuery
     }
 
@@ -733,7 +733,7 @@ public struct Query<T>: Encodable, Equatable where T: ParseObject {
      */
     public func select(_ keys: [String]) -> Query<T> {
         var mutableQuery = self
-        mutableQuery.keys = keys.joined(separator: ",")
+        mutableQuery.keys = keys
         return mutableQuery
     }
 
--- a/Sources/ParseSwift/Types/Query.swift
+++ b/Sources/ParseSwift/Types/Query.swift
@@ -697,6 +697,16 @@ public struct Query<T>: Encodable, Equatable where T: ParseObject {
 
     /**
      Exclude specific keys for a `ParseObject`. Default is to nil.
+     - parameter keys: A variadic list of keys include in the result.
+     */
+    public func exclude(_ keys: String...) -> Query<T> {
+        var mutableQuery = self
+        mutableQuery.excludeKeys = keys
+        return mutableQuery
+    }
+
+    /**
+     Exclude specific keys for a `ParseObject`. Default is to nil.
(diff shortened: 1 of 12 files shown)


cbaker6:
@dplewis there's a problem here that you may know the answer to quickly. Query `select` and `exclude` are not working as expected as the Parse Server is returning all keys. It almost seems like it's ignoring the arrays I send for `keys` and `excludeKeys`, respectively.

My guess at the problem has something to do with the use of `POST` instead of `GET` and putting everything related to the query (besides the endpoint) in the `body`. When posting the query, it includes `_method = "GET"`, which I believe overrides the `POST` at the server level based on https://github.com/parse-community/parse-server/blob/e6ac3b6932186a6a85601c65143305f86f63e948/src/middlewares.js#L336-L343

All other queries (first, hint, explain, etc.) I've tested work using `POST` for query. When using `select` with a simple query, `var query2 = GameScore.query().select(["yolo"])`, what gets sent in the body to the server is

```
["_method": "GET", "skip": 0, "limit": 100, "where": [:], "keys": ["yolo"]]
``` 

Note that, using `find` which sets `limit = 1` works. Am I doing something wrong here? From what I see on the server-side, https://github.com/parse-community/parse-server/blob/e53b6c2f87b209e5b6a459f64111995caee3b7b9/src/RestQuery.js#L100-L116 

when using `select`, `keys` should be at the same level as `limit`. When it comes to `exclude`, `excludeKeys` should be at the same level as `limit` as well. Is that correct?

cbaker6:
@dplewis @pmmlo I think found the issue and it's looking to be server side. It also looks like the test cases written for this is missing a corner case. I'll update with my reasoning after some more testing

cbaker6:
False alarm, I still don’t know where the issue is at

pmmlo:
Thanks for looking into this @cbaker6  Is there any documentation on how requests should be constructed for the server?  Otherwise, I could take a look at one of the other SDKs.

cbaker6:
No documentation that I know of besides how to us it in JS SDK, which which is I tried to create the current implementation from. I couldn't find this in the REST documentation. The JS SDK has it.

The iOS SDK has `selectedKeys`, but it looks like it's doing a standard query and then stripping the non selected keys on the client side.

cbaker6:
@pmmlo the last commit I made works by using a manual join of the array, implying that the Parse server doesn't like they way `keys` and `excludeKeys` are being encoded.

The test failures are do them still looking for arrays instead of strings. If you want something that works, you can use that commit until I figure out what the actual issue is

cbaker6:
@dplewis @mtrezza so I believe my false alarm mentioned earlier was actually a real one and the problem is server side.

In short, the `keys` and `excludeKeys` on the server are not digesting actual JSON arrays, `["yolo", ...]`, but instead digesting `"yolo", ...`. Somehow JS treats both of these the same (I say this because the fixed PR I have for the server works with just removing 1 test case, more on this in the PR), but the server doesn't like it the way the Swift JSON encoder sends it, `["yolo", ...]`. Note that this is the JSON encoder straight from Apples Swift source code and not the custom one used here for `ParseObjects`, so if it was encoding arrays incorrectly I'm sure it would have been caught awhile ago.

I'll submit the PR now so you the changes. I'm still running one more additional test locally to verify somethings

cbaker6:
Done with my local tests with this framework and the server changes and they work. It looks like the tests pass on the server side and on this client SDK. 

Of course, if there's a test on the JS SDK that used `select([])`, it will break if the server side PR is accepted. I'm guessing the other `select` and `exclude` tests will pass though

cbaker6:
@TomWFox can you review the doc comments I added? There was a missing instance method for `exclude`.

cbaker6:
@pmmlo If you use docker and need a working server with this now, you can use my image, `netreconlab/parse-server:latest` https://hub.docker.com/repository/docker/netreconlab/parse-server

Otherwise, you will have to see what comes of the discussion and PR on the server repo

pmmlo:
> If you use docker and need a working server with this now, you can use my image, `netreconlab/parse-server:latest` https://hub.docker.com/repository/docker/netreconlab/parse-server

Thanks for all the contribs @cbaker6   I am still playing around with Parse to see if it would be good for personal use.  Really solid work!

cbaker6:
@pmmlo thanks for discovering the bug! Definitely let us know if you find anything else.

I did find a little bit of documentation on `keys` and `excludeKeys`, but it highlights the linked bug fix and the examples can be improved https://docs.parseplatform.org/rest/guide/#query-constraints

Sources/ParseSwift/Types/Query.swift
```diff
@@ -697,9 +697,21 @@ public struct Query<T>: Encodable, Equatable where T: ParseObject {
...
+     */
+    public func exclude(_ keys: String...) -> Query<T> {
+        var mutableQuery = self
+        mutableQuery.excludeKeys = keys
+        return mutableQuery
+    }
+
+    /**
+     Exclude specific keys for a `ParseObject`. Default is to nil.
+     - parameter keys: An arrays of keys to exclude.
```
TomWFox, line 711:
```suggestion
     - parameter keys: An array of keys to exclude.
```

PR #89, opened 2021-03-08
Add batch transactions

cbaker6:
Support for batch transactions https://github.com/parse-community/parse-server/pull/5849

- [x] Add transaction for `saveAll` and `deleteAll`
- [x] Add `batchLimit` argument to all publisher (SwiftUI) `saveAll` and `deleteAll` (this was an oversight when adding publishers) 
- [x] Update documentation
- [x] Add test cases 
- [x] Update changelog
- [x] Switch default playgrounds to macOS. Add notes to playground files about when to switch
- [x] Improve playground `Pointer` example by querying with `include` and `includeAll`
- [x] Only send idempotent on POST and PUT 
- [x] Prepare for new release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,11 @@
 [Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.1.6...main)
 * _Contributing to this repo? Add info about your change here to be included in the next release_
 
+### 1.1.7
+[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.1.6...1.1.7)
+
 __New features__
+- Add transaction support to batch saveAll and deleteAll ([#89](https://github.com/parse-community/Parse-Swift/pull/89)), thanks to [Corey Baker](https://github.com/cbaker6).
 - Add modifiers to containsString, hasPrefix, hasSuffix ([#85](https://github.com/parse-community/Parse-Swift/pull/85)), thanks to [Corey Baker](https://github.com/cbaker6).
 
 __Improvements__
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Parse-Swift Changelog
 
 ### main
-[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.1.6...main)
+[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.1.7...main)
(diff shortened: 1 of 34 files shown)


cbaker6:
@TomWFox can you look through the documentation when you get a chance?

cbaker6:
@davimacedo @dplewis the way to leverage server level transactions from the client side seemed to be intuitive when I took a look. Where it seemed to get tricky (from your discussions on the server and JS SDK) is when handling unsaved child objects and possibly running into batch limits? Is there anything I missed or should be on the lookout for?

One possible issue I can see coming up is when unsaved child objects are involved, the saving of those children are essentially mini transactions, meaning that if batch1 of children save correctly, but batch2 (say the parent objects) fails, transaction 1 was still committed to the DB. This may confuse the developer because not be thinking in terms of children/parent. This essentially acts "almost" like a regular batch without transactions. Is this correct? If so, I don't see it as much of a problem and can't really think of a workaround on the server level (seems you need to cache some transaction list or something similar). It might be easier to do this on the client side though. Basically, if the client hits an error in batch2, they call deleteAll for batch1. Though, they might need the `masterKey` if batch1 contained files.

Question about batch limits, I've seen in the iOS and Android SDKs a batch limit of 50, but see the JS mention 100. The JS SDK seems to have a way to change this. Parse-Swift has a way to change it as well. Is there a real sever limit or was that left over from Parse.com? If there is a limit, I assumed this can be configured by the developer, and hinted towards this in the docs.

One of things I did was change the batch limit to whatever the transaction size is: https://github.com/parse-community/Parse-Swift/blob/8939e47495033ce036da7d403cc86eedd14b43b7/Sources/ParseSwift/Objects/ParseObject.swift#L123-L128

I also added this warning in the docs: https://github.com/parse-community/Parse-Swift/blob/8939e47495033ce036da7d403cc86eedd14b43b7/Sources/ParseSwift/Objects/ParseObject.swift#L71-L73

Let me know what you think and I'm interested in hearing any limitations/problems you see.

cbaker6:
Noting a comment from @davimacedo in the linked PR:

> Hi guys. I am working on this (to include the validations and the tests) but I found out a bug in Parse Server when deleting objects in a transaction. I am working now on a first PR to Parse Server before fixing this one here.

If this bug still exists, I can remove transactions for `deleteAll` for now if there's not an easy fix

PR #97, opened 2021-03-19
added Twitter and Facebook login

abs8090:
This PR adds Twitter and Facebook login via the ParseAuthenticatable protocol. 
The user of ParseTwitter or ParseFacebook still needs to implement TwitterKit or FBSDKCoreKit to obtain authentication keys and pass them to ParseTwitter.logn() or ParseFacebook.login() to authenticate with ParseServer.


cbaker6:
Thanks @abs8090 for your contribution! I will look over soon.

In the mean time can you add an entry to the CHANGELOG.md file, something like:

```
__New features__
- Add ParseTwitter and ParseFacebook authentication ([#97](https://github.com/parse-community/Parse-Swift/pull/97)), thanks to [YOUR NAME](https://github.com/abs8090).
```

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
```diff
@@ -0,0 +1,251 @@
...
+    /// Authentication keys required for Facebook authentication.
+    enum AuthenticationKeys: String, Codable {
+        case id // swiftlint:disable:this identifier_name
+        case token
+
+        /// Properly makes an authData dictionary with the required keys.
+        /// - parameter facebookId: Required id for the user.
+        /// - parameter authToken: Required identity token for the user.
+        /// - returns: Required authData dictionary.
+        func makeDictionary(facebookId: String,
```
cbaker6, line 31: Can you change these keys to match facebooks API? `facebookId` is okay, but `authToken` should be `accessToken`. I'm assuming you left this over from the `ParseApple` copy. Luckily, the newer key for limited login called `authenticationToken` described https://github.com/parse-community/parse-server/pull/7219 which uses the same key for the server, `token`. This means you should change all `authToken` to `authenticationToken`. You will need to propagate this change through the rest of ParseFacebook, comments, and the test cases.

This means, you also need to add a new field to your code for `accessToken` is this is used for Facebooks graph API. It looks the developer should either provide `accessToken` or `token`, but not both at the same time.

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
```diff
@@ -0,0 +1,251 @@
...
+ Provides utility functions for working with Facebook User Authentication and `ParseUser`'s.
+ Be sure your Parse Server is configured for [sign in with Facebook](https://docs.parseplatform.org/parse-server/guide/#configuring-parse-server-for-sign-in-with-facebook).
+ For information on acquiring Facebook sign-in credentials to use with `ParseFacebook`, refer to [Facebook's Documentation](https://developers.facebook.com/docs/facebook-login/limited-login.
+ */
+public struct ParseFacebook<AuthenticatedUser: ParseUser>: ParseAuthentication {
+
+    /// Authentication keys required for Facebook authentication.
+    enum AuthenticationKeys: String, Codable {
+        case id // swiftlint:disable:this identifier_name
+        case token
```
cbaker6, line 25: You can leave `token`, but this is for limited login. You need to also add the original login `case accessToken`.

You can add coding keys right after the case statements since SwiftLint won't like `access_token`:

```
enum CodingKeys: String, CodingKey {
            case id // swiftlint:disable:this identifier_name 
            case accessToken = "access_token"
}
```

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
```diff
@@ -0,0 +1,251 @@
...
+             AuthenticationKeys.token.rawValue: authToken]
+        }
+
+        /// Verifies all mandatory keys are in authData.
+        /// - parameter authData: Dictionary containing key/values.
+        /// - returns: `true` if all the mandatory keys are present, `false` otherwise.
+        func verifyMandatoryKeys(authData: [String: String]?) -> Bool {
+            guard let authData = authData,
+                  authData[AuthenticationKeys.id.rawValue] != nil,
+                  authData[AuthenticationKeys.token.rawValue] != nil else {
```
cbaker6, line 44: note that this key check will change because of what I mentioned earlier. `id` is required, but either `token` or `accessToken` is required, not both. Should only return `false` if both are missing.

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
```diff
@@ -0,0 +1,251 @@
...
+     Login a `ParseUser` *asynchronously* using Facebook authentication.
+     - parameter withFacebookId: The `Facebook userId` from `FacebookSDK`.
+     - parameter authToken: The `authToken` from `FacebookSDK`.
+     - parameter options: A set of header options sent to the server. Defaults to an empty set.
+     - parameter callbackQueue: The queue to return to after completion. Default value of .main.
+     - parameter completion: The block to execute.
+     */
+
+    func login(withFacebookId: String,
+               authToken: String,
```
cbaker6, line 68: `withFacebookId` should be `facebookId`. 

After you change `authToken`, you should make a copy of this function, but it should take `accessToken` instead of `authenticationToken`. The documentation comments above should describe is the method is for limited login or not. This should be done for the sync, async, and publisher methods

cbaker6 (changes requested):

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
cbaker6: From the docs, https://docs.parseplatform.org/parse-server/guide/#facebook-authdata it looks like Facebook also takes an `expirationData` which seems it should be a required field (I need to verify). You should add this and the coding keys will look like:

```swift
enum CodingKeys: String, CodingKey { 
  case id // swiftlint:disable:this identifier_name 
  case accessToken = "access_token" 
  case expirationData = "expiration_date"
}
```

Sources/ParseSwift/Authentication/3rd Party/ParseTwitter.swift
```diff
@@ -0,0 +1,267 @@
...
+ Provides utility functions for working with Twitter User Authentication and `ParseUser`'s.
+ Be sure your Parse Server is configured for [sign in with Twitter](https://docs.parseplatform.org/parse-server/guide/#configuring-parse-server-for-sign-in-with-twitter).
+ For information on acquiring Twitter sign-in credentials to use with `ParseTwitter`, refer to [Twitter's Documentation](https://developer.twitter.com/en/docs/authentication/guides/log-in-with-twitter.
+ */
+public struct ParseTwitter<AuthenticatedUser: ParseUser>: ParseAuthentication {
+
+    /// Authentication keys required for Twitter authentication.
+    enum AuthenticationKeys: String, Codable {
+        case id // swiftlint:disable:this identifier_name
+        case token
```
cbaker6, line 26: The keys for twitter don't seem correct, the [documentation]() shows:

```javascript
"twitter": {
    "id": "user's Twitter id number as a string",
    "consumer_key": "your application's consumer key",
    "consumer_secret": "your application's consumer secret",
    "auth_token": "an authorized Twitter token for the user with your application",
    "auth_token_secret": "the secret associated with the auth_token"
  }
``` 

which will need coding keys:

```swift
enum CodingKeys: String, CodingKey { 
  case id // swiftlint:disable:this identifier_name 
  case consumer_key = "consumer_key"
  case consumer_secret = "consumer_secret"
  case authToken = "auth_token"
  case authTokenSecret = "auth_token_secret"  
}
```

Sources/ParseSwift/Authentication/3rd Party/ParseTwitter.swift
```diff
@@ -0,0 +1,267 @@
...
+     - parameter authToken: The Twitter `authToken` obtained from Twitter SDK for the user.
+     - parameter authTokenSecret: The Twitter `authSecretToken` obtained from Twitter SDK for the user
+     - parameter options: A set of header options sent to the server. Defaults to an empty set.
+     - parameter callbackQueue: The queue to return to after completion. Default value of .main.
+     - parameter completion: The block to execute.
+     */
+
+    func login(withTwitterId: String,
+               username: String,
+               authToken: String,
```
cbaker6, line 77: Need to modify the methods to take the keys I mentioned above, also, remove `with` from `withTwitterId`. Be sure to do this for all of the sync, async, and publisher methods

cbaker6 (changes requested):

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
cbaker6: You will need to convert the `expirationDate` to the correct version of string. After the `enum AuthenticationKeys` declaration in ParseFacebook add the following computed property:

```swift
static let dateFormatter: DateFormatter = {
    var dateFormatter = DateFormatter()
    dateFormatter.locale = Locale(identifier: "en_US_POSIX")
    dateFormatter.timeZone = TimeZone.init(secondsFromGMT: 0)
    dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
    return dateFormatter
}()
```

Your `makeDictionary` function will look something like:

```swift
func makeDictionary(facebookId: String, accessToken: String? = nil, authenticationToken: String? = nil, expirationDate: Date) throws -> [String: String]? {

  let dateString = Self.dateFormatter.string(from: expirationDate)
  var returnDictionary = [AuthenticationKeys.id.rawValue: facebookId,
              AuthenticationKeys.expirationDate.rawValue: dateString]
  if let accessToken = accessToken {
    returnDictionary[AuthenticationKeys. accessToken.rawValue] = accessToken 
  }  else if let authenticationToken = authenticationToken {
    returnDictionary[AuthenticationKeys.token.rawValue] = authenticationToken
  } else {
    throw ParseError(code: .unknownError,
                    message: "Missing required key. Need to provide \"accessToken\" or \"authenticationToken\"")
  }
  return returnDictionary
}

```

Sources/ParseSwift/Authentication/3rd Party/ParseTwitter.swift
cbaker6: For twitter, your method will look like:

```swift
func login(twitterId: String,
                screenName: String,
                authToken: String,
                authTokenSecret: String,
                consumerKey: String,
                consumerSecret: String,
                options: API.Options = [],
                completion: @escaping (Result<AuthenticatedUser, ParseError>) -> Void) {
```

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
```diff
@@ -0,0 +1,395 @@
...
+        case authenticationToken
+        case accessToken
+        case expirationDate
+        
+        enum CodingKeys: String, CodingKey {
+          case id // swiftlint:disable:this identifier_name
+          case token
+          case accessToken = "access_token"
+          case expirationDate = "expiration_date"
+        }
```
cbaker6, line 34: Can you add a space after this line

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
```diff
@@ -0,0 +1,395 @@
...
+        /// Properly makes an authData dictionary with the required keys.
+        /// - parameter userId: Required id for the user.
+        /// - parameter authenticationToken: Required identity token for the user for Facebook limited login.
+        /// - parameter accessToken: Required identity token for the user for Facebook graph API
+        /// - parameter expirationDate: required expiration data for user authentication for Facebook login
```
cbaker6, line 39:
```suggestion
        /// - parameter expirationDate: Required expiration data for user authentication for Facebook login
```

cbaker6 (changes requested):

Sources/ParseSwift/Authentication/3rd Party/ParseTwitter.swift
```diff
@@ -0,0 +1,313 @@
...
+          case screenName  = "screen_name"
+        }
+        /// Properly makes an authData dictionary with the required keys.
+        /// - parameter twitterId: Required id for the user.
+        /// - parameter screenName: The `Twitter screenName` from `Twitter session`.
+        /// - parameter consumerKey: The `Twitter consumerKey` from `Twitter SDK`
+        /// - parameter consumerSecret: The `Twitter consumerSecret` from `Twitter SDK`..
+        /// - parameter authToken: Required Twitter authToken obtained from Twitter SDK for the user.
+        /// - parameter authTokenSecret: Required Twitter authSecretToken obtained from Twitter SDK for the user.
+        /// - returns: Required authData dictionary.
```
cbaker6, line 47:
```suggestion
        /// - returns: authData dictionary.
```

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
```diff
@@ -0,0 +1,395 @@
...
+        /// - returns: Required authData dictionary.
```
cbaker6, line 40:
```suggestion
        /// - returns: authData dictionary.
```

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
```diff
@@ -0,0 +1,395 @@
...
+        func makeDictionary(userId: String?, accessToken: String?, authenticationToken: String?, expirationDate: Date?) throws -> [String: String]? {
+            
+            let validAuthenticationToken = authenticationToken != nil || accessToken != nil
```
cbaker6, line 43: `validAuthenticationToken` doesn’t seem needed. Can you look at the version of `makeDictionary` I provided to you in the comments. This method doesn’t need to through and can be smaller

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
```diff
@@ -0,0 +1,395 @@
...
+        /// - parameter authData: Dictionary containing key/values.
+        /// - returns: `true` if all the mandatory keys are present, `false` otherwise.
+        func verifyMandatoryKeys(authData: [String: String]?) -> Bool {
+            guard let authData = authData,
+                  authData[AuthenticationKeys.id.rawValue] != nil,
+                  authData[AuthenticationKeys.expirationDate.rawValue] != nil else {
+                return false
+            }
+            
+            if let _ = authData[AuthenticationKeys.accessToken.rawValue] {
```
cbaker6, line 73: Can you check the warnings Swiftlint is giving you in Xcode? The way these lines are written and the spacing, I’m sure there’s a number of warnings that need to be addressed. There should only be 1 warning in the SDK and it has a comment saying to leave it

cbaker6 (changes requested):

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
```diff
@@ -0,0 +1,395 @@
...
+    /**
+     Login a `ParseUser` *asynchronously* using Facebook authentication for graph API login.
+     - parameter userId: The `Facebook userId` from `FacebookSDK`.
+     - parameter accessToken: The `accessToken` from `FacebookSDK`.
+     - parameter options: A set of header options sent to the server. Defaults to an empty set.
+     - parameter callbackQueue: The queue to return to after completion. Default value of .main.
+     - parameter completion: The block to execute.
+     */
+    func login(userId: String,
+               accessToken: String?,
```
cbaker6, line 126:
```suggestion
               accessToken: String,
```

cbaker6: If a developer is using this method it is “mandatory” they specify `accessToken`. I made this change in two places, but you should make the change in the rest.

cbaker6 (changes requested):

cbaker6:
I updated your branch with the latest main, you should should `pull` your branch and then proceed with the rest of the changes

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
abs8090: can you please specify after which line I need to add the space? Also, do you want me to add a new line or just white space?

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
abs8090: keys have been changed so they match Facebook's API, they are: `id`, `expirationDate`, `accessToken` and `authenticationToken`

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
abs8090: `CodingKeys` have been added following your suggestions. 
Additionally, the `makeDictionary` function has been modified and now `userId` and `expirationDate` are mandatory, but `accessToken` and `authenticationToken` are optional.

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
abs8090: done

Sources/ParseSwift/Authentication/3rd Party/ParseTwitter.swift
abs8090: done

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
cbaker6: It was a newline that was needed. When you add functions, you should separate them by 1 newline

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
cbaker6: I resolved this

cbaker6:
The PR is looking good!

You had some inconsistencies in ParseTwitter with, `twitterId`, `userId`, and `user`. I switched them all to `userId` as it looks like that's what Twitter uses.

Things you need to do:

- [ ] Install SwiftLint from the [contributors](https://github.com/parse-community/Parse-Swift/blob/main/CONTRIBUTING.md) guide and fix the warnings.
- [ ] Verify you have all of the necessary keys for Facebook and Twitter and make sure your testcases are checking what they need to check

cbaker6:
This looks ready to go!

Thanks for making your first contribution!

cbaker6 (approved):

PR #101, opened 2021-03-22
Fix allowCustomObjectId

cbaker6:
- [x] Fixed a bug when the `allowCustomObjectId` config was enabled that was adding the objectId to the url endpoint
- [x] When `allowCustomObjectId == true`, the SDK will throw an error if `objectId` is missing
- [x] Doc nits


cbaker6:
@TomWFox do you think we should leave this open for a bit and see if we find any other nits in the docs?

http://parseplatform.org/Parse-Swift/api/index.html

TomWFox:
Totally up to you!

PR #107, opened 2021-03-30
Authentication Provider Delegate

Vortec4800:
This is an early draft PR for new Authentication Delegate functionality.

The purpose of this is to allow custom 3rd party auth providers to restore their internal state when Parse loads and restores from serialized data. An example would be to load tokens into an OAuth SDK, or some SSO SDK, so you can run authenticated calls on those SDKs with an existing user. It also gives the delegate a chance to tell Parse that the auth data provided is no longer valid, and shouldn't be used to restore this user.

I wasn't 100% sure about the best places to add these hooks, so I took my best stab. Let me know if there are better locations to do things like notify the auth providers to restore, or saving the auth provider info to the user and keychain.

I modified the LDAP provider as an example, even though it doesn't actually have use for it. This is mostly to show how it would work, we can either do the same thing on the other providers so they are a good example for future providers that might need this, or it can be removed from the LDAP as it's not needed.

Also, take particular note of the LDAP file line 77 where I set the `authProviderType` to see if that's the best time to do it. If it is, I can add that same function to the other login/link calls.

Finally, let me know if there are any naming conventions or other changes you'd like to see.

### DIFF
--- a/Sources/ParseSwift/Authentication/3rd Party/ParseAuthenticationRegistry.swift
+++ b/Sources/ParseSwift/Authentication/3rd Party/ParseAuthenticationRegistry.swift
@@ -1,0 +1,60 @@
+//
+//  ParseAuthenticationRegistry.swift
+//  ParseSwift
+//
+//  Created by Cory Imdieke on 3/29/21.
+//  Copyright © 2021 Parse Community. All rights reserved.
+//
+
+import Foundation
+
+public protocol ParseAuthenticationDelegate {
+	/**
+	 Called when restoring third party authentication credentials that have been serialized,
+	 such as session keys, user id, etc.
+	 - This method will be executed on a background thread.
+	 - parameter authData: The auth data for the provider. This value may be `nil` when unlinking an account.
+	 - returns `true` - if the `authData` was succesfully synchronized,
+	 or `false` if user should not longer be associated because of bad `authData`.
+	 */
+	func restoreAuthentication(withAuthData authData: [String: String]) -> Bool
+}
+
+/**
+ This is a system that maintains a list of auth providers available to the SDK.
+ The primary reason is so we can notify these providers when we're going to restore an
+ auth session that originated from that provider. This gives the provider a chance to
+ re-instate any data necessary for use in the app.
+ */
+public struct ParseAuthenticationRegistry {
+	private static var authRegistry: [String: ParseAuthenticationDelegate] = [:]
+	private static var restoredAuthResults: [String: Bool] = [:]
+
+	/**
+	Registers an authentication delegate for a specific auth type. Optional, only necessary if your custom authentication
+	provider needs to respond to the authentication delegate methods and restore state during user restoration.
+	*/
+	public static func registerAuthenticationDelegate(_ delegate: ParseAuthenticationDelegate, forAuthType type: String) {
+		// TODO: We have a check to make sure each type is only registered once, but do we actually care? Should we just let this succeed?
+		assert(authRegistry[type] == nil, "Tried to register an auth delegate to a type name that already exists")
+
+		authRegistry[type] = delegate
+	}
+
+	internal static func restoreAuthentication(forType type: String, authData: [String: String]?) -> Bool {
+		guard let authData = authData else {
+			return true
+		}
+
+		if let previousResponse = restoredAuthResults[type] {
+			// Already restored, return same result as last time
+			return previousResponse
+		} else if let delegate = authRegistry[type] {
+			let result = delegate.restoreAuthentication(withAuthData: authData)
+			restoredAuthResults[type] = result
+			return result
+		} else {
+			return true
+		}
+	}
+}
--- a/Sources/ParseSwift/Authentication/3rd Party/ParseLDAP.swift
+++ b/Sources/ParseSwift/Authentication/3rd Party/ParseLDAP.swift
@@ -48,7 +48,9 @@ public struct ParseLDAP<AuthenticatedUser: ParseUser>: ParseAuthentication {
     public static var __type: String { // swiftlint:disable:this identifier_name
         "ldap"
     }
-    public init() { }
+    public init() {
+		ParseAuthenticationRegistry.registerAuthenticationDelegate(self, forAuthType: Self.__type)
+	}
 }
 
 // MARK: Login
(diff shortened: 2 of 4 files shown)


Vortec4800:
@cbaker6 take a look, also note this doesn't address any of the error stuff we've been talking about, we can deal with that once we put this portion to bed.

cbaker6:
I read through the code and your description. I don't understand what this is used for. Maybe if you add test cases it will help clarify? The testcases should not on test the code added, but should show scenarios of how this is being used. You can look at the LDAP testcases to see how to go about this, https://github.com/parse-community/Parse-Swift/blob/main/Tests/ParseSwiftTests/ParseLDAPTests.swift

> The purpose of this is to allow custom 3rd party auth providers to restore their internal state when Parse loads and restores from serialized data. An example would be to load tokens into an OAuth SDK, or some SSO SDK, so you can run authenticated calls on those SDKs with an existing user. It also gives the delegate a chance to tell Parse that the auth data provided is no longer valid, and shouldn't be used to restore this user.

What does "restore user" mean? When would this happen in the current life cycle of a user? Is the being called every time the app is opened?

Vortec4800:
The delegate method gets called when the user is first loaded from the keychain, so on or near launch when the SDK is first accessed. Here is a copy of the delegate call from my custom provider, so you can get an idea of the functionality performed here:

```swift
	public func restoreAuthentication(withAuthData authData: [String : String]) -> Bool {
		let hasAccessToken = authData["access_token"] != nil
		let hasExpiryDate = authData["expire_date"] != nil
		
		guard hasAccessToken && hasExpiryDate else {
			return false
		}
		
		guard let userID = authData["id"], let credentialData64 = authData["encoded_credential"], let credentialJSON = Data(base64Encoded: credentialData64), let credential = try? JSONDecoder().decode(OAuthSwiftCredential.self, from: credentialJSON) else {
			// Unable to reconstitute credential, don't let this user log in
			return false
		}
		
		// Restore on client
		self.userID = userID
		self.oauthSwift.client = OAuthSwiftClient(credential: credential)
		print("Using valid auth token")
		return true
	}
```

Sources/ParseSwift/Authentication/3rd Party/ParseLDAP.swift
```diff
@@ -200,6 +207,13 @@ public extension ParseLDAP {
     #endif
 }
 
+extension ParseLDAP: ParseAuthenticationDelegate {
+	public func restoreAuthentication(withAuthData authData: [String: String]) -> Bool {
```
cbaker6, line 211: This should be a requirement in `ParseAuthentication` like https://github.com/parse-community/Parse-Swift/blob/3852a93ed00e532c6214086700fb98e5eb4bd03f/Sources/ParseSwift/Authentication/Protocols/ParseAuthentication.swift#L42-L52

and then have the convenience/default implementation that you have. Similar to: https://github.com/parse-community/Parse-Swift/blob/3852a93ed00e532c6214086700fb98e5eb4bd03f/Sources/ParseSwift/Authentication/Protocols/ParseAuthentication.swift#L164-L169

When an auth type wants to change from the default, they will implement it

Sources/ParseSwift/Authentication/3rd Party/ParseAuthenticationRegistry.swift
```diff
@@ -0,0 +1,60 @@
...
+	private static var authRegistry: [String: ParseAuthenticationDelegate] = [:]
+	private static var restoredAuthResults: [String: Bool] = [:]
+
+	/**
+	Registers an authentication delegate for a specific auth type. Optional, only necessary if your custom authentication
+	provider needs to respond to the authentication delegate methods and restore state during user restoration.
+	*/
+	public static func registerAuthenticationDelegate(_ delegate: ParseAuthenticationDelegate, forAuthType type: String) {
+		// TODO: We have a check to make sure each type is only registered once, but do we actually care? Should we just let this succeed?
+		assert(authRegistry[type] == nil, "Tried to register an auth delegate to a type name that already exists")
```
cbaker6, line 39: This should probably throw a created ParseError instead

Sources/ParseSwift/Authentication/3rd Party/ParseAuthenticationRegistry.swift
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ This is a system that maintains a list of auth providers available to the SDK.
+ The primary reason is so we can notify these providers when we're going to restore an
+ auth session that originated from that provider. This gives the provider a chance to
+ re-instate any data necessary for use in the app.
+ */
+public struct ParseAuthenticationRegistry {
+	private static var authRegistry: [String: ParseAuthenticationDelegate] = [:]
+	private static var restoredAuthResults: [String: Bool] = [:]
```
cbaker6, line 31: You can initialize these in this way, `private static var restoredAuthResults = [String: Bool]()`, same with the one above

cbaker6 (changes requested):

cbaker6:
Do you have swift lint as specified in the contributors guide? https://github.com/parse-community/Parse-Swift/blob/main/CONTRIBUTING.md. If so, be sure address any warnings. There should only be 1 warning in the SDK and it has a comment with my name above.

Can you add testcases for this? We won't be able to accept it with out the necessary testcases.

cbaker6:
For your examples and testing you should remove the changes you made to LDAP since you said LDAP wouldn't use it and instead use the `TestAuth` here: https://github.com/parse-community/Parse-Swift/blob/3852a93ed00e532c6214086700fb98e5eb4bd03f/Tests/ParseSwiftTests/ParseAuthenticationTests.swift#L63-L102

You can add your tests to the test file above.

Vortec4800:
Just wanted to let you know that I've been working on the app that would use this more, and I think it would be best to abandon this PR. I think the need for this was more important in the original ObjC SDK, but now the same behavior can be achieved pretty easily without needing all of this extra code complexity.

Thanks for working through this with me in the meantime.

cbaker6:
@Vortec4800 this is great to hear! Are you able to prove an example of how you can achieve the same behavior with the new SDK?

Vortec4800:
Sure thing. It's embarrassingly simple given the size of this PR...

On launch, immediately after the SDK is initialized, we check to see if we have a current user and then call the method we need to:

```swift
if let authData = AppUser.current?.authData?["graph"] ?? nil {
	_ = self.authHelper.restoreAuthentication(withAuthData: authData)
}
```

That's... it. The only other time a delegate might get called is during logout/login, but you can call whatever you need to set up during that process as well, as your code is the one calling the login method anyway.

Issue #93, opened 2021-03-14
Parse Installation


novemTeam:
I have an issue where the installation always prints an error. 
This is my code:

```
@main
struct MyApp: App {
    
    init() {
        ParseSwift.initialize(applicationId: "id", clientKey: "key", serverURL: URL(string: "https://myapp.back4app.io")!)
        DispatchQueue.main.async {
            Installation.current?.save { results in
                switch results {
                case .success(_):
                    print("[Novem] Succesufully saved Installation to ParseServer")
                case .failure(let error):
                    print("[Novem] Failed to update installation: \(error)")
                }
            }
        }
    }
    var body: some Scene {
        WindowGroup {
            MainView()
        }
    }
}
```

I get this error every time my app launches

[...]

PR #112, opened 2021-04-02
Recreate new installation after deletion from keychain

### LINKED ISSUE #110: ParseInstallation.current is always nil after ParseUser.logout
When `ParseUser.logout` is called it also clears the current ParseInstallation by calling `ParseInstallation.deleteCurrentContainerFromKeychain`. This is expected.

https://github.com/parse-community/Parse-Swift/blob/f3f22444e7b15fd2d248bfb4fd1244edb76ed748/Sources/ParseSwift/Objects/ParseUser.swift#L80-L84

https://github.com/parse-community/Parse-Swift/blob/f3f22444e7b15fd2d248bfb4fd1244edb76ed748/Sources/ParseSwift/Objects/ParseInstallation.swift#L185-L190

Afterwards `ParseInstallation.current` always returns nil.

Expected: A new `ParseInstallation` is created when `ParseInstallation.currentInstallationContainer.currentInstallation` is nil.

I think this could be the right place to create a new installation when its missing:

https://github.com/parse-community/Parse-Swift/blob/f3f22444e7b15fd2d248bfb4fd1244edb76ed748/Sources/ParseSwift/Objects/ParseInstallation.swift#L136-L142

I currently see no way to create a new `ParseInstallation` object from outside the framework. The only place in code I found which actively sets a new installation is `ParseInstallation.updateKeychainIfNeeded`. This only gets called on `fetch`, `save` or similar requests. However I can not call `fetch` or `save` when `ParseInstallation.current` is nil.

cbaker6:
Automatically create a new installation after it's deleted from the Keychain. A deletion typically occurs on user logout. Fix #110 

- [x] add changelog entry
- [x] add tests

### DIFF
--- a/Tests/ParseSwiftTests/ParseUserCombineTests.swift
+++ b/Tests/ParseSwiftTests/ParseUserCombineTests.swift
@@ -293,32 +293,42 @@ class ParseUserCombineTests: XCTestCase { // swiftlint:disable:this type_body_le
         }
 
         let expectation1 = XCTestExpectation(description: "Logout user1")
-        let publisher = User.logoutPublisher()
-            .sink(receiveCompletion: { result in
+        DispatchQueue.main.async {
+            let oldInstallationId = BaseParseInstallation.current?.installationId
+            let publisher = User.logoutPublisher()
+                .sink(receiveCompletion: { result in
 
-                if case let .failure(error) = result {
-                    XCTFail(error.localizedDescription)
+                    if case let .failure(error) = result {
+                        XCTFail(error.localizedDescription)
+                    }
+                    expectation1.fulfill()
+
+            }, receiveValue: { _ in
+                if let userFromKeychain = BaseParseUser.current {
+                    XCTFail("\(userFromKeychain) wasn't deleted from Keychain during logout")
                 }
-                expectation1.fulfill()
+                DispatchQueue.main.async {
+                    if let installationFromMemory: CurrentInstallationContainer<BaseParseInstallation>
+                        = try? ParseStorage.shared.get(valueFor: ParseStorage.Keys.currentInstallation) {
+                        if installationFromMemory.installationId == oldInstallationId {
+                            // swiftlint:disable:next line_length
+                            XCTFail("\(installationFromMemory) wasn't deleted and recreated in memory during logout")
+                        }
+                    }
 
-        }, receiveValue: { _ in
-            if let userFromKeychain = BaseParseUser.current {
-                XCTFail("\(userFromKeychain) wasn't deleted from Keychain during logout")
-            }
-
-            if let installationFromMemory: CurrentInstallationContainer<BaseParseInstallation>
-                = try? ParseStorage.shared.get(valueFor: ParseStorage.Keys.currentInstallation) {
-                XCTFail("\(installationFromMemory) wasn't deleted from memory during logout")
-            }
-
-            #if !os(Linux) && !os(Android)
-            if let installationFromKeychain: CurrentInstallationContainer<BaseParseInstallation>
-                = try? KeychainStore.shared.get(valueFor: ParseStorage.Keys.currentInstallation) {
-                XCTFail("\(installationFromKeychain) wasn't deleted from Keychain during logout")
-            }
-            #endif
-        })
-        publisher.store(in: &subscriptions)
+                    #if !os(Linux) && !os(Android)
+                    if let installationFromKeychain: CurrentInstallationContainer<BaseParseInstallation>
+                        = try? KeychainStore.shared.get(valueFor: ParseStorage.Keys.currentInstallation) {
+                        if installationFromKeychain.installationId == oldInstallationId {
+                            // swiftlint:disable:next line_length
+                            XCTFail("\(installationFromKeychain) wasn't deleted and recreated in Keychain during logout")
+                        }
+                    }
+                    #endif
+                }
+            })
+            publisher.store(in: &subscriptions)
+        }
         wait(for: [expectation1], timeout: 20.0)
     }
 
--- a/Tests/ParseSwiftTests/ParseUserCombineTests.swift
+++ b/Tests/ParseSwiftTests/ParseUserCombineTests.swift
@@ -294,7 +294,11 @@ class ParseUserCombineTests: XCTestCase { // swiftlint:disable:this type_body_le
 
         let expectation1 = XCTestExpectation(description: "Logout user1")
         DispatchQueue.main.async {
-            let oldInstallationId = BaseParseInstallation.current?.installationId
+            guard let oldInstallationId = BaseParseInstallation.current?.installationId else {
+                XCTFail("Should have unwrapped")
+                expectation1.fulfill()
+                return
+            }
             let publisher = User.logoutPublisher()
                 .sink(receiveCompletion: { result in
 
@@ -310,8 +314,8 @@ class ParseUserCombineTests: XCTestCase { // swiftlint:disable:this type_body_le
                 DispatchQueue.main.async {
                     if let installationFromMemory: CurrentInstallationContainer<BaseParseInstallation>
                         = try? ParseStorage.shared.get(valueFor: ParseStorage.Keys.currentInstallation) {
-                        if installationFromMemory.installationId == oldInstallationId {
-                            // swiftlint:disable:next line_length
+                        if installationFromMemory.installationId == oldInstallationId
+                            && installationFromMemory.installationId != nil {
                             XCTFail("\(installationFromMemory) wasn't deleted and recreated in memory during logout")
                         }
                     }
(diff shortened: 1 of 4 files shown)


cbaker6:
@funkenstrahlen can you try out this branch and see if you run into the same issue?

The testcases pass and new installationIds are created

Tests/ParseSwiftTests/ParseUserTests.swift
```diff
@@ -1083,59 +1083,75 @@ class ParseUserTests: XCTestCase { // swiftlint:disable:this type_body_length
...
-                #if !os(Linux) && !os(Android)
-                if let installationFromKeychain: CurrentInstallationContainer<BaseParseInstallation>
-                    = try? KeychainStore.shared.get(valueFor: ParseStorage.Keys.currentInstallation) {
-                    XCTFail("\(installationFromKeychain) wasn't deleted from Keychain during logout")
-                }
-                #endif
+                        #if !os(Linux) && !os(Android)
+                        if let installationFromKeychain: CurrentInstallationContainer<BaseParseInstallation>
+                            = try? KeychainStore.shared.get(valueFor: ParseStorage.Keys.currentInstallation) {
+                            if installationFromKeychain.installationId == oldInstallationId {
```
funkenstrahlen: What do you think about extending the test case here and also checking if `installationFromKeychain.currentInstallation != nil`?

Tests/ParseSwiftTests/ParseUserTests.swift
```diff
@@ -1083,59 +1083,75 @@ class ParseUserTests: XCTestCase { // swiftlint:disable:this type_body_length
...
-                    XCTFail("\(installationFromMemory) wasn't deleted from memory during logout")
-                }
+                case .success:
+                    if let userFromKeychain = BaseParseUser.current {
+                        XCTFail("\(userFromKeychain) wasn't deleted from Keychain during logout")
+                    }
+                    DispatchQueue.main.async {
+                        if let installationFromMemory: CurrentInstallationContainer<BaseParseInstallation>
+                            = try? ParseStorage.shared.get(valueFor: ParseStorage.Keys.currentInstallation) {
+                            if installationFromMemory.installationId == oldInstallationId {
```
funkenstrahlen: What do you think about extending the test case here and also checking if `installationFromKeychain.currentInstallation != nil`?

Tests/ParseSwiftTests/ParseUserTests.swift
```diff
@@ -1083,59 +1083,75 @@ class ParseUserTests: XCTestCase { // swiftlint:disable:this type_body_length
...
             do {
-                let encoded = try ParseCoding.jsonEncoder().encode(logoutResponse)
-                return MockURLResponse(data: encoded, statusCode: 200, delay: 0.0)
+                try User.logout()
+                if let userFromKeychain = BaseParseUser.current {
+                    XCTFail("\(userFromKeychain) wasn't deleted from Keychain during logout")
+                }
+                DispatchQueue.main.async {
+                    if let installationFromKeychain = BaseParseInstallation.current {
+                        if installationFromKeychain.installationId == oldInstallationId {
```
funkenstrahlen: What do you think about extending the test case here and also checking if `installationFromKeychain.currentInstallation != nil`?

Tests/ParseSwiftTests/ParseUserCombineTests.swift
```diff
@@ -293,32 +293,42 @@ class ParseUserCombineTests: XCTestCase { // swiftlint:disable:this type_body_le
...
-                = try? KeychainStore.shared.get(valueFor: ParseStorage.Keys.currentInstallation) {
-                XCTFail("\(installationFromKeychain) wasn't deleted from Keychain during logout")
-            }
-            #endif
-        })
-        publisher.store(in: &subscriptions)
+                    #if !os(Linux) && !os(Android)
+                    if let installationFromKeychain: CurrentInstallationContainer<BaseParseInstallation>
+                        = try? KeychainStore.shared.get(valueFor: ParseStorage.Keys.currentInstallation) {
+                        if installationFromKeychain.installationId == oldInstallationId {
```
funkenstrahlen: What do you think about extending the test case here and also checking if `installationFromKeychain.currentInstallation != nil`?

Tests/ParseSwiftTests/ParseUserCombineTests.swift
```diff
@@ -293,32 +293,42 @@ class ParseUserCombineTests: XCTestCase { // swiftlint:disable:this type_body_le
...
-                XCTFail("\(installationFromMemory) wasn't deleted from memory during logout")
-            }
+            }, receiveValue: { _ in
+                if let userFromKeychain = BaseParseUser.current {
+                    XCTFail("\(userFromKeychain) wasn't deleted from Keychain during logout")
+                }
+                DispatchQueue.main.async {
+                    if let installationFromMemory: CurrentInstallationContainer<BaseParseInstallation>
+                        = try? ParseStorage.shared.get(valueFor: ParseStorage.Keys.currentInstallation) {
+                        if installationFromMemory.installationId == oldInstallationId {
```
funkenstrahlen: What do you think about extending the test case here and also checking if `installationFromKeychain.currentInstallation != nil`?

Tests/ParseSwiftTests/ParseUserTests.swift
cbaker6: This may not always be the case by the time the test gets here because of the dispatching to main queue. If it is nil when it gets here, it won't be in a couple of seconds

funkenstrahlen:
Thank you for quick response and the PR! I added some suggestions above to have a more detailed test suite.

I added this branch to my project and can confirm it works a lot better now! Thank you!

However I still need to manually save the newly created installation to the server after logout because every subsequent `ParseInstallation.fetch` fails with `ParseError code=-1 error=Cannot fetch an object without id` before it has been saved to the server once.

```swift
    func logout() {
        PodliveUser.logout { (result) in
            switch result {
            case .failure(let error):
                print("Could not logout: " + error.localizedDescription)
            case .success():
                print("successfully logged out")
                NotificationCenter.default.post(name: Notification.Name.userUpdated, object: nil)
                PodliveInstallation.current?.save(completion: { (result) in
                    switch result {
                    case .success(_):
                        self.loginAsAnonymousUser()
                    case .failure(let error):
                        print("Could not save installation: " + error.localizedDescription)
                    }
                })
            }
        }
    }
```

Do you think saving the installation to the server should be done by the framework after logout?

Tests/ParseSwiftTests/ParseUserTests.swift
funkenstrahlen: You're right. I haven't thought about the dispatch. The additional test case is not required. Has just been a thought as it's not checked right now.

cbaker6: #113 should provide more predictable behavior. Let me know your thoughts on that thread

cbaker6:
> Do you think saving the installation to the server should be done by the framework after logout?

I believe this comment may answer this question: https://github.com/parse-community/Parse-Swift/issues/93#issuecomment-798935361

cbaker6:
> 

> Thank you for quick response and the PR! I added some suggestions above to have a more detailed test suite.
> 
> I added this branch to my project and can confirm it works a lot better now! Thank you!
> 
> However I still need to manually save the newly created installation to the server after logout because every subsequent `ParseInstallation.fetch` fails with `ParseError code=-1 error=Cannot fetch an object without id` before it has been saved to the server once.
> 
> ```swift
>     func logout() {
>         PodliveUser.logout { (result) in
>             switch result {
>             case .failure(let error):
>                 print("Could not logout: " + error.localizedDescription)
>             case .success():
>                 print("successfully logged out")
>                 NotificationCenter.default.post(name: Notification.Name.userUpdated, object: nil)
>                 PodliveInstallation.current?.save(completion: { (result) in
>                     switch result {
>                     case .success(_):
>                         self.loginAsAnonymousUser()
>                     case .failure(let error):
>                         print("Could not save installation: " + error.localizedDescription)
>                     }
>                 })
>             }
>         }
>     }
> ```
> 
> Do you think saving the installation to the server should be done by the framework after logout?

Since you logout/login immediately, I can see the possibility of odd behavior because of the need for Installation to be on the main queue. There can be times your anonymous login occurs before the new installation is created. You can minimize this odd behavior by logging in anonymously after a short time, say a second (dispatch after a second). The changes discussed in #113 should help with this also and you might not need to wait anytime after to log back in, but with the current implementation, you should wait a second or ask the user to press a button to login

funkenstrahlen:
> > Do you think saving the installation to the server should be done by the framework after logout?
> 
> I believe this comment may answer this question: [#93 (comment)](https://github.com/parse-community/Parse-Swift/issues/93#issuecomment-798935361)

Thank you for linking the related issue thread. Actually I do have custom properties in my Installation object and therefore need to retrieve it manually. However I think you are absolutely correct that fetching the Installation at this point should not be part of the framework.

funkenstrahlen:
> Since you logout/login immediately, I can see the possibility of odd behavior because of the need for Installation to be on the main queue. There can be times your anonymous login occurs before the new installation is created. You can minimize this odd behavior by logging in anonymously after a short time, say a second (dispatch after a second).

That's correct. Great you thought about that! I think implementing #113 is the best approach to create a more predictable behavior.

cbaker6:
> That's correct. Great you thought about that! I think implementing #113 is the best approach to create a more predictable behavior.

Even with the changes, you should probably implement one of my suggestions I mentioned above. You can see that because of threading the new installation may not be ready immediately in the keychain: https://github.com/parse-community/Parse-Swift/blob/b671243fa650672af182f0e36e0d947a38d8f3f9/Tests/ParseSwiftTests/ParseUserCombineTests.swift#L313-L336

PR #114, opened 2021-04-02
ParseInstallation no longer requires dispatching to main queue

### LINKED ISSUE #113: Remove automatic retrieval of badge in ParseInstallation
Getting the badge value automatically is something native to the Objective-c SDK (and currently Parse-Swift). It requires using UIKit or AppKit to retrieve and accessing it properly means `ParseInstallation` has to always be called on the main queue. This leads to awkward implementation and usage of `ParseInstallation`. In addition, Parse-Swift works on OSs that don't have badges like Linux and Android.

I plan to remove the automatic retrieval of badge. If a developer wants to populate this field they can create their own dispatches before adding it to `ParseInstallation`.

This will lead to easier usage and less buggy and more predictable (example https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1568) behavior when using `ParseInstallation`.

Feel free to comment with your thoughts...

cbaker6:
Reasoning/discussion in #113. This is a breaking changing, but will allow for easier and more predictable use of `ParseInstallation`. If the developer wants to save the latest badge value, they should retrieve it before saving the installation.

Fix #113 

- [x] Remove the usage of UIKit and AppKit in the SDK.
- [x] Removed dispatching when using `ParseInstallation`
- [x] Updated playgrounds 
- [x] Added changelog entry
- [x] Updated Installation documentation 
- [x] Prepare for new release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,14 @@
 # Parse-Swift Changelog
 
 ### main
-[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.2.6...main)
+[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.3.0...main)
 * _Contributing to this repo? Add info about your change here to be included in the next release_
+
+### 1.3.0
+[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.2.6...1.3.0)
+
+__Improvements__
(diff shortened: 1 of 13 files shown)


cbaker6:
@TomWFox can you look over the updated docs for ParseInstallation when you get a chance?

TomWFox (approved):
LGTM 👍

PR #148, opened 2021-05-22
Add ParseAnalytics

### LINKED ISSUE #147: Cannot use [object Object] as a query parameter
I wrote the following code: 

```
static func getAuthor(from book: Book, completion: @escaping ((Result<Author, Error>) -> Void)) {
    let query = Author.query().include("book").where("book" == book)
    query.first { results in
        switch results {
            case .success(let author):
                completion(.success(author))
            case .failure(let error):
                completion(.failure(error))
        }
    }
}
```

And expected to be able to fetch the Author that had the property "book" of type Pointer<Book> and of value of the book I passed in. (I previously fetched the Book and passed it directly into here)

Here is my Author:

```
struct Author: ParseObject {

    var objectId: String?
    var createdAt: Date?
    var updatedAt: Date?
    var ACL: ParseACL?

    var book: Book?
}
```

But when the query runs I get the error ParseError(code: ParseSwift.ParseError.Code.invalidJSON, message: "You cannot use [object Object] as a query parameter.")

Note: Sorry if this by any chance possible through a different approach (and please let me know how :)) The parse sdk allowed me to do this with where(equalTo:)... so I was assuming this was a bug.

cbaker6:
Adds ParseAnalytics to track events. For iOS 14.0, macOS 11.0, macCatalyst 14.0, tvOS 14.0, you
 will need to request tracking authorization for ParseAnalytics to work.
 See Apple's [documentation](https://developer.apple.com/documentation/apptrackingtransparency) for more for details.

- [x] Add analytics
- [x] Add test cases
- [x] Add playground examples
- [x] Add entry to changelog
- [x] Add == ParseObject to Query Close #147 
- [X] Fixed an bug when deleting a ParseFile. It was accidentally downloading the file while deleting
- [x] Make ParseFile, ParseCloud, ParseRelation, Pointer `Hashable` 
- [x] Prepare for release


Vortec4800:
I just found this and was surprised to see App Tracking Transparency is required for this. Does ParseAnalytics use the advertiser identifier or some other protected identifier?

PR #155, opened 2021-06-05
Append instead of replace when using query

cbaker6:
Allow appending of select, exclude, include, and fields

- [x] Add test cases
- [x] Add changelog


dplewis:
Can you teach me swift? Lol

cbaker6:
Anytime you are ready, your JS skills should transfer over seamlessly to Swift

PR #158, opened 2021-06-11
Don't use transactions by default for internal SDK calls

### LINKED ISSUE #157: Decoding error when saving nested object
Hi all,

I am encountering an issue when saving a nested object.

I have a teacher object that has a name (string) and a language (custom object)

```
struct teacher : ParseObject {
    
    var objectId: String?
    var createdAt: Date?
    var updatedAt: Date?
    var ACL: ParseACL?
    
    var name : String
    var langauge : language
}
```

```
struct language : ParseObject{
    var objectId: String?
    var createdAt: Date?
    var updatedAt: Date?
    var ACL: ParseACL?
    
    var code : String
}
```

I have no problem creating and saving language objects, but when I try to save a Teacher object I am getting a decoding error:

ParseError(code: ParseSwift.ParseError.Code.unknownError, message: "decoding error: typeMismatch(Swift.Array<Any>, Swift.DecodingError.Context(codingPath: [], debugDescription: \"Expected to decode Array<Any> but found a dictionary instead.\", underlyingError: nil))")

I am running the parse server on a localhost. macOS Big Sur and SwiftUI.

I have also tried saving the example of Author and Book in the Swift Playgrounds example code but I am getting exactly the same output.

Any one would know how can I solve this? Or what I am missing?

Kind regards,
Pau

cbaker6:
Using transactions with the latest release of parse-server using mongoDB produces errors while saving. Because of this, transactions are disabled by default. If the developer wants to use transactions internally (was used for saveAll children) in the SDK, they should set `useTransactionsInternally=true` when configuring the client.

Mongo users should update to Swift SDK 1.8.1 as versions 1.2.1 - 1.8.0 use transactions by default.

Note: transactions seem to work fine with postgres.

- [x] Disable automatic use of transactions internally
- [x] Add `useTransactionsInternally` to `ParseConfiguration`
- [x] Update playgrounds
- [x] Add changelog
- [x] Prepare for release   

Close #157

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,17 @@
 # Parse-Swift Changelog
 
 ### main
-[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.8.0...main)
+[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.8.1...main)
 * _Contributing to this repo? Add info about your change here to be included in the next release_
+
+### 1.8.0
+[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.8.0...1.8.1)
 
 __Improvements__
 - Append instead of replace when using query select, exclude, include, and fields ([#155](https://github.com/parse-community/Parse-Swift/pull/155)), thanks to [Corey Baker](https://github.com/cbaker6).
+
+__Fixes__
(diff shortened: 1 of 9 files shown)


cbaker6:
@davimacedo @dplewis is this why https://github.com/parse-community/Parse-SDK-JS/pull/1090 isn't merged?

PR #165, opened 2021-06-16
new: allow for custom error codes from Parse Cloud

### LINKED ISSUE #164: Parse Swift can't decode custom ParseCloud errors
Throwing a custom error code from a Parse Cloud function results in:

```
ParseError(code: ParseSwift.ParseError.Code.unknownError, message: "Error decoding parse-server response: Optional(<NSHTTPURLResponse: 0x100d74140> { URL: http://localhost:1337/1/jobs/test } { Status Code: 200, Headers {\n    \"Content-Type\" =     (\n        \"application/json\"\n    );\n} }) with error: The data couldn’t be read because it is missing. 
```

Failing case:

```
    func testCustomError() {

        let encoded: Data = "{\"error\":\"Error: Custom Error\",\"code\":2000}".data(using: .utf8)!

        MockURLProtocol.mockRequests { _ in
            return MockURLResponse(data: encoded, statusCode: 200, delay: 0.0)
        }
        do {
            let cloud = Cloud(functionJobName: "test")
            _ = try cloud.startJob()
            XCTFail("Should have thrown ParseError")
        } catch {
            if let error = error as? ParseError {
                XCTAssertNotEqual(error.code, ParseError.Code.unknownError)
                XCTAssertEqual(error.code.rawValue, 2000)
            } else {
                XCTFail("Should have thrown ParseError")
            }
        }
    }
```

Proposed solution:

Add `.other` enum to error, and `intCode` property to Parse.Error

dblythy:
Closes #164

@cbaker6 let me know your thoughts!

### DIFF
--- a/Sources/ParseSwift/API/URLSession+extensions.swift
+++ b/Sources/ParseSwift/API/URLSession+extensions.swift
@@ -40,13 +40,9 @@ extension URLSession {
                 return try .success(mapper(responseData))
             } catch {
                 if var error = try? ParseCoding.jsonDecoder().decode(ParseError.self, from: responseData) {
-                    do {
-                        let json = try JSONSerialization.jsonObject(with: responseData, options: []) as? [String : Any]
-                        if let code = json?["code"] as? Int {
-                            error.intCode = code
-                        }
-                    } catch {
-                        /* */
+                    if let json = try? JSONSerialization.jsonObject(with: responseData, options: []) as? [String: Any],
+                       let code = json["code"] as? Int {
+                        error.intCode = code
                     }
                     return .failure(error)
                 }
--- a/Sources/ParseSwift/API/URLSession+extensions.swift
+++ b/Sources/ParseSwift/API/URLSession+extensions.swift
@@ -39,7 +39,15 @@ extension URLSession {
             do {
                 return try .success(mapper(responseData))
             } catch {
-                if let error = try? ParseCoding.jsonDecoder().decode(ParseError.self, from: responseData) {
+                if var error = try? ParseCoding.jsonDecoder().decode(ParseError.self, from: responseData) {
+                    do {
+                        let json = try JSONSerialization.jsonObject(with: responseData, options: []) as? [String : Any]
+                        if let code = json?["code"] as? Int {
+                            error.intCode = code
+                        }
+                    } catch {
+                        /* */
+                    }
                     return .failure(error)
                 }
                 guard let parseError = error as? ParseError else {
--- a/Sources/ParseSwift/Types/ParseError.swift
+++ b/Sources/ParseSwift/Types/ParseError.swift
@@ -15,7 +15,7 @@ public struct ParseError: ParseType, Decodable, Swift.Error {
     public let code: Code
     public let message: String
     public var intCode: Int? = nil
-    init(code: Code, message: String) {
+    public init(code: Code, message: String) {
         self.code = code
         self.message = message
         self.intCode = code.rawValue
--- a/Sources/ParseSwift/Types/ParseError.swift
+++ b/Sources/ParseSwift/Types/ParseError.swift
@@ -14,7 +14,7 @@ import Foundation
 public struct ParseError: ParseType, Decodable, Swift.Error {
     public let code: Code
     public let message: String
-    public var intCode: Int? = nil
+    public var intCode: Int?
     public init(code: Code, message: String) {
         self.code = code
         self.message = message
@@ -353,7 +353,7 @@ public struct ParseError: ParseType, Decodable, Swift.Error {
          a non-2XX status code.
          */
         case xDomainRequest = 602
-        
+
         /**
          Error code indicating any other custom error sent from Parse Cloud
          */
--- a/Sources/ParseSwift/Types/ParseError.swift
+++ b/Sources/ParseSwift/Types/ParseError.swift
@@ -15,7 +15,7 @@ public struct ParseError: ParseType, Decodable, Swift.Error {
     public let code: Code
     public let message: String
     public var intCode: Int?
-    public init(code: Code, message: String) {
+    init(code: Code, message: String) {
         self.code = code
         self.message = message
         self.intCode = code.rawValue
--- a/Sources/ParseSwift/Types/ParseError.swift
+++ b/Sources/ParseSwift/Types/ParseError.swift
@@ -14,6 +14,12 @@ import Foundation
 public struct ParseError: ParseType, Decodable, Swift.Error {
     public let code: Code
     public let message: String
+    public var intCode: Int? = nil
+    init(code: Code, message: String) {
+        self.code = code
+        self.message = message
+        self.intCode = code.rawValue
+    }
 
     /// A textual representation of this error.
     public var localizedDescription: String {
@@ -347,6 +353,12 @@ public struct ParseError: ParseType, Decodable, Swift.Error {
          a non-2XX status code.
          */
         case xDomainRequest = 602
+        
+        /**
+         Error code indicating any other custom error sent from Parse Cloud
+         */
+
+        case other
     }
 }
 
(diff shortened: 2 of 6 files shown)


cbaker6:
@dblythy thanks for your PR! Here’s a response I gave to a similar proposal? https://github.com/parse-community/Parse-Swift/pull/106#issuecomment-809542815

From your addition, does this mean the other SDKs currently allow users to create their own errors and mask them as Parse errors? I could be wrong here, but I would think errors would come from a set list of known Parse errors. If a user generates an error, maybe it can fall under “unknownError” or add a new error code on the server specific to user errors? Not sure what’s the best approach, but definitely open to hear you and the servers team thoughts.

My fear is allowing developers to mask their errors as parse errors and then people submit issues to Parse repos claiming they are Parse errors when they are not

cbaker6:
My reasoning of using unknownError is because of how Flo documented ParseError before I took over it: https://github.com/parse-community/Parse-Swift/blob/603486ae6f2d27a7d5fdf9f94379bf045265ad36/Sources/ParseSwift/Types/ParseError.swift#L35-L37

It seems believed all SDK errors, technically custom SDK errors should fall under this error. So I can see other errors going here as well

Sources/ParseSwift/Types/ParseError.swift
```diff
@@ -14,6 +14,12 @@ import Foundation
 public struct ParseError: ParseType, Decodable, Swift.Error {
     public let code: Code
     public let message: String
+    public var intCode: Int? = nil
+    init(code: Code, message: String) {
```
cbaker6, line 18: I’m on my phone, but probably will need to make the init public for devs to use it outside of the SDK

Sources/ParseSwift/API/URLSession+extensions.swift
```diff
@@ -39,7 +39,15 @@ extension URLSession {
...
             } catch {
-                if let error = try? ParseCoding.jsonDecoder().decode(ParseError.self, from: responseData) {
+                if var error = try? ParseCoding.jsonDecoder().decode(ParseError.self, from: responseData) {
+                    do {
+                        let json = try JSONSerialization.jsonObject(with: responseData, options: []) as? [String : Any]
+                        if let code = json?["code"] as? Int {
+                            error.intCode = code
+                        }
+                    } catch {
+                        /* */
```
cbaker6, line 49: Was something suppose to go here?

cbaker6 (changes requested):

Sources/ParseSwift/API/URLSession+extensions.swift
dblythy: I'm not sure. I'm not very experienced with swift and not sure how the linting works. In JS, the Lint gets mad when you have empty `catch` blocks.

cbaker6: Did you install swiftlint? https://github.com/parse-community/Parse-Swift/blob/main/CONTRIBUTING.md

Is it giving you any warnings? There should only be 1 warning in the SDK that shows up in < Xcode 12.5 and is marked with a comment by me. All others should be fixed before merging

cbaker6: This block should be written different, the empty catch won’t work

dblythy: I haven't got swiftlint working. I'll work on it.

The block was to get "code" out of the response payload. I couldn't work out how else to do it (I don't actually write in Swift, this work is for a colleague.

cbaker6:
@dblythy I made some updates to better fit the SDK. I'll make some comments on the rest

Tests/ParseSwiftTests/ParseCloudTests.swift
```diff
@@ -356,6 +356,28 @@ class ParseCloudTests: XCTestCase { // swiftlint:disable:this type_body_length
         }
     }
 
+    func testCustomError() {
+
+        let encoded: Data = "{\"error\":\"Error: Custom Error\",\"code\":2000}".data(using: .utf8)!
```
cbaker6, line 361: Can you change this to a `guard` statement? Currently, we don't force unwrap in the SDK

Sources/ParseSwift/Types/ParseError.swift
```diff
@@ -12,12 +12,25 @@ import Foundation
  An object with a Parse code and message.
  */
 public struct ParseError: ParseType, Decodable, Swift.Error {
+    /// The value representing the error from the Parse Server.
     public let code: Code
+    /// The text representing the error from the Parse Server.
     public let message: String
+    /// An error value representing a custom error from the Parse Server.
+    public let otherCode: Int?
+    public init(code: Code, message: String) {
```
cbaker6, line 21: Based on my comment earlier https://github.com/parse-community/Parse-Swift/issues/164#issuecomment-862269923

I still don't see the need for developers to create errors on the client-side and call it a "Parse Error". If a developer creates an error in CloudCode, it should decode with the current additions you have in this PR and doesn't need this init

cbaker6 (changes requested):

cbaker6:
@TomWFox can you checkout the added documentation?

Sources/ParseSwift/Types/ParseError.swift
```diff
@@ -347,14 +360,25 @@ public struct ParseError: ParseType, Decodable, Swift.Error {
          a non-2XX status code.
          */
         case xDomainRequest = 602
+
+        /**
+         Error code indicating any other custom error sent from Parse Cloud
```
TomWFox, line 365:
```suggestion
         Error code indicating any other custom error sent from the Parse Server.
```

Sources/ParseSwift/Types/ParseError.swift
```diff
@@ -18,7 +18,7 @@ public struct ParseError: ParseType, Decodable, Swift.Error {
     public let message: String
     /// An error value representing a custom error from the Parse Server.
     public let otherCode: Int?
-    public init(code: Code, message: String) {
+    init(code: Code, message: String) {
         self.code = code
         self.message = message
         self.otherCode = code.rawValue
```
cbaker6, line 24:
```suggestion
        self.otherCode = nil
```

Tests/ParseSwiftTests/ParseErrorTests.swift
```diff
@@ -0,0 +1,64 @@
...
+        #if !os(Linux) && !os(Android)
+        try KeychainStore.shared.deleteAll()
+        #endif
+        try ParseStorage.shared.deleteAll()
+    }
+
+    func testEncode() throws {
+        let code = -1
+        let message = "testing ParseError"
+        guard let encoded: Data = "{\"error\":\"\(message)\",\"code\":\(code)}".data(using: .utf8) else {
```
cbaker6, line 40:
```suggestion
        guard let encoded = "{\"error\":\"\(message)\",\"code\":\(code)}".data(using: .utf8) else {
```

Tests/ParseSwiftTests/ParseErrorTests.swift
```diff
@@ -0,0 +1,64 @@
...
+        let decoded = try ParseCoding.jsonDecoder().decode(ParseError.self, from: encoded)
+        XCTAssertEqual(decoded.code.rawValue, code)
+        XCTAssertEqual(decoded.message, message)
+        XCTAssertEqual(decoded.localizedDescription, "ParseError code=\(code) error=\(message)")
+    }
+
+    func testEncodeOther() throws {
+        let code = 2000
+        let message = "testing ParseError"
+        guard let encoded: Data = "{\"error\":\"\(message)\",\"code\":\(code)}".data(using: .utf8) else {
```
cbaker6, line 53:
```suggestion
        guard let encoded = "{\"error\":\"\(message)\",\"code\":\(code)}".data(using: .utf8) else {
```

cbaker6:
@dblythy looking good, can you add a changelog entry?

cbaker6 (approved):

PR #191, opened 2021-07-11
Save all fields of User and Installation to keychain

cbaker6:
Previously, `ParseUser.current` and `ParseInstallation.current` only persisted the required Parse properties to the Keychain and didn't contain any developer added properties. This was done to keep the Keychain as small as possible, but required developers to have to fetch the `ParseUser` and `ParseInstallation` to access any added properties after app restart.

This update removes those restrictions and persists all `ParseUser` and `ParseInstallation` properties to the Keychain. A side effect of this is Keychains will now be larger than they were previously, so developers should be cautious of what types of data they store to `ParseUser` and `ParseInstallation`.


funkenstrahlen:
❤️

PR #202, opened 2021-07-29
fix: use correct auth keys for Facebook and Twitter login

### LINKED ISSUE #201: ParseFacebook sets accessToken instead of access_token
Hi!

Unless I'm doing something wrong, I'm unable to login with Facebook.

The issue is cause by the fact that Parse Server uses `access_token`, yet Parse Swift uses `accessToken`.

Failing test:

```
func testMakeDictionary() {
        let authData = ParseFacebook<User>
            .AuthenticationKeys.id.makeDictionary(userId: "testing",
                                                  accessToken: "testAccessToken",
                                                  authenticationToken: "testAuthToken")
        XCTAssertEqual(authData["id"],"testing")
        XCTAssertEqual(authData["access_token"],"testAccessToken")

        let secondAuthData = ParseFacebook<User>
            .AuthenticationKeys.id.makeDictionary(userId: "testing",
                                                  accessToken: nil,
                                                  authenticationToken: "testAuthToken")

        XCTAssertEqual(secondAuthData["id"],"testing")
        XCTAssertEqual(secondAuthData["token"],"testAuthToken")
    }
```

https://github.com/parse-community/parse-server/blob/9ea355b4635226ae4da17c8cc5fb0321e3fdec5e/src/Adapters/Auth/facebook.js#L24

dblythy:
Closes #201.

Sets correct payload for FB login.

### DIFF
--- a/Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
+++ b/Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
@@ -22,16 +22,9 @@ public struct ParseFacebook<AuthenticatedUser: ParseUser>: ParseAuthentication {
     /// Authentication keys required for Facebook authentication.
     enum AuthenticationKeys: String, Codable {
         case id // swiftlint:disable:this identifier_name
-        case authenticationToken
-        case accessToken
-        case expirationDate
-
-        enum CodingKeys: String, CodingKey { // swiftlint:disable:this nesting
-          case id // swiftlint:disable:this identifier_name
-          case authenticationToken = "token"
-          case accessToken = "access_token"
-          case expirationDate = "expiration_date"
-        }
+        case authenticationToken = "token"
+        case accessToken = "access_token"
+        case expirationDate = "expiration_date"
 
         /// Properly makes an authData dictionary with the required keys.
         /// - parameter userId: Required id for the user.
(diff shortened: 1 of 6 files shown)


cbaker6:
Though this will work, it uncovers a deeper issue. The Keys should be encoded instead of using the coding keys directly

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
```diff
@@ -44,19 +44,19 @@ public struct ParseFacebook<AuthenticatedUser: ParseUser>: ParseAuthentication {
                             authenticationToken: String?,
                             expiresIn: Int? = nil) -> [String: String] {
 
-            var returnDictionary = [AuthenticationKeys.id.rawValue: userId]
+            var returnDictionary = [CodingKeys.id.rawValue: userId]
```
cbaker6, line 47: Please revert all of these back to AuthenticationKeys

cbaker6: The actual fix should be applied to lines 25-27:

```swift
enum AuthenticationKeys: String, Codable {
        case id // swiftlint:disable:this identifier_name
        case authenticationToken = "token"
        case accessToken = "access_token"
        case expirationDate = "expiration_date"
```

You can deleted the CodingKeys, 29-34

Tests/ParseSwiftTests/ParseFacebookCombineTests.swift
```diff
@@ -84,7 +84,22 @@ class ParseFacebookCombineTests: XCTestCase { // swiftlint:disable:this type_bod
         #endif
         try ParseStorage.shared.deleteAll()
     }
+    func testMakeDictionary() {
```
cbaker6, line 87: Can you move this test to ParseFacebookCombineTests

cbaker6 (changes requested):

Sources/ParseSwift/Authentication/3rd Party/ParseFacebook.swift
cbaker6: Can you also make the similar change to Twitter: https://github.com/parse-community/Parse-Swift/blob/252df53eee6e8f95178eb3f58ce7124ab5673b3a/Sources/ParseSwift/Authentication/3rd%20Party/ParseTwitter.swift#L27-L39

Tests/ParseSwiftTests/ParseFacebookCombineTests.swift
dblythy: I ended up removing this because it's already covered by test (just the keys are wrong)

cbaker6:
Can you add a change log entry?

cbaker6 (approved):

cbaker6:
LGTM thanks for the fixes!

dblythy:
Thank you @cbaker6!

PR #208, opened 2021-08-02
Removing random Int in the reconnection interval of ParseLiveQuery and added warning in playgrounds

lsmilek1:
Added warning in playgrounds related to:
- mixed environment with both custom and server generated objectId not supported
- ParseObject.fetch() returns empty object instead of .objectNotFound ParseError

Removing random Int in the reconnection interval of ParseLiveQuery:
- If socket is being open by user, there is no reason for delay -> Added boolean check

### DIFF
--- a/ParseSwift.playground/Pages/1 - Your first Object.xcplaygroundpage/Contents.swift
+++ b/ParseSwift.playground/Pages/1 - Your first Object.xcplaygroundpage/Contents.swift
@@ -239,6 +239,11 @@ let scoreToFetch = GameScore(objectId: savedScore?.objectId)
 
 //: Asynchronously (preferred way) fetch this GameScore based on it's objectId alone.
 scoreToFetch.fetch { result in
+    /*: Warning: server does return empty object {} instead of a ParseServer error `.objectNotFount`
+        if the object does not exist yet. This might result in decoding the result into a empty
+        struct. User `query.first()` of `query.find()` instead if your would like to recieve
+        and handle an `.objectNotFount` error
+    */
     switch result {
     case .success(let fetchedScore):
         print("Successfully fetched: \(fetchedScore)")
--- a/ParseSwift.playground/Pages/15 - Custom ObjectId.xcplaygroundpage/Contents.swift
+++ b/ParseSwift.playground/Pages/15 - Custom ObjectId.xcplaygroundpage/Contents.swift
@@ -18,6 +18,11 @@ npm start -- --appId applicationId --clientKey clientKey --masterKey masterKey -
 /*: In Xcode, make sure you are building the "ParseSwift (macOS)" framework.
  */
 
+/*: Warning: A mixed environment with both custom and server generated ObjectId
+    is not supported. SDK throws error if set to use custom objectId
+    but any object without defined objectId should be saved
+*/
+
 initializeParseCustomObjectId()
 
 //: Create your own value typed `ParseObject`.
@@ -89,6 +94,11 @@ score.fetch { result in
     switch result {
     case .success(let fetchedScore):
         print("Successfully fetched: \(fetchedScore)")
+    /*: Warning: server does return empty object {} instead of a ParseServer error `.objectNotFount`
+        if the object does not exist yet. This might result in decoding the result into a empty
+        struct. User `query.first()` of `query.find()` instead if your would like to recieve
+        and handle an `.objectNotFount` error
+    */
     case .failure(let error):
         assertionFailure("Error fetching: \(error)")
     }
--- a/Sources/ParseSwift/LiveQuery/ParseLiveQuery.swift
+++ b/Sources/ParseSwift/LiveQuery/ParseLiveQuery.swift
@@ -223,7 +223,7 @@ extension ParseLiveQuery {
 
     var reconnectInterval: Int {
         let min = NSDecimalNumber(decimal: Swift.min(30, pow(2, attempts) - 1))
-        return Int.random(in: 0 ..< Int(truncating: min))
+        return Int(truncating: min)
     }
 
     func resumeTask(completion: @escaping (Error?) -> Void) {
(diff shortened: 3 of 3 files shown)


Sources/ParseSwift/LiveQuery/ParseLiveQuery.swift
```diff
@@ -223,7 +223,7 @@ extension ParseLiveQuery {
 
     var reconnectInterval: Int {
         let min = NSDecimalNumber(decimal: Swift.min(30, pow(2, attempts) - 1))
-        return Int.random(in: 0 ..< Int(truncating: min))
+        return Int(truncating: min)
```
cbaker6, line 226: What's the reason for changing this?

ParseSwift.playground/Pages/1 - Your first Object.xcplaygroundpage/Contents.swift
```diff
@@ -239,6 +239,11 @@ let scoreToFetch = GameScore(objectId: savedScore?.objectId)
 
 //: Asynchronously (preferred way) fetch this GameScore based on it's objectId alone.
 scoreToFetch.fetch { result in
+    /*: Warning: server does return empty object {} instead of a ParseServer error `.objectNotFount`
```
cbaker6, line 242: nits, change `.objectNotFount` to `.objectNotFound`

ParseSwift.playground/Pages/15 - Custom ObjectId.xcplaygroundpage/Contents.swift
```diff
@@ -89,6 +94,11 @@ score.fetch { result in
     switch result {
     case .success(let fetchedScore):
         print("Successfully fetched: \(fetchedScore)")
+    /*: Warning: server does return empty object {} instead of a ParseServer error `.objectNotFount`
```
cbaker6, line 97: same nits here

cbaker6 (changes requested):

Sources/ParseSwift/LiveQuery/ParseLiveQuery.swift
lsmilek1: As I mentioned [here](https://community.parseplatform.org/t/parseswift-sdk-observe-livequery-websocket-status/1739/35), I suspect a code race as the reconnectionInterval is mostly 0 for first two attempts and `resumeTask()` id deallocating the delegate with `URLSession.liveQuery.delegates.removeValue(forKey: self.task)` 

https://github.com/parse-community/Parse-Swift/blob/e3f51f4d5f1520d694edac248ebb883f2c05d375/Sources/ParseSwift/LiveQuery/ParseLiveQuery.swift#L229-L249

Is there a reason for the random Int? Because without the LiveQuery reconnection seems to behave much more reproducible and is even reconnection on server reset without holding the execution with a breakpoint.

cbaker6: > As I mentioned here, I suspect a code race as the reconnectionInterval is mostly 0 for first two attempts and resumeTask() is deallocating the delegate with URLSession.liveQuery.delegates.removeValue(forKey: self.task)

This will need more justification and discussion before a change like this can be accepted (you can open an [issue](https://github.com/parse-community/Parse-Swift/issues) to begin that discussion. The first few attempts should come out to zero because it's a random increasing interval that increases based on the amount of attempts, so zero the first 2-3 attempts makes sense and represents trying to reconnect immediately after disconnection.

> Because without the random Int the LiveQuery reconnection seems to behave much more reproducible and is even reconnecting on server reset without holding the execution with a breakpoint.

This seems odd because your change makes the value `0`, but you mention the first few values from the `reconnectionInterval` are `0` which produces the same thing.

> Is there a reason for the random Int?

In networking you generally have some form of increasing backoff when attempting to reconnect.

I recommend changing this back to the original is even if there is a problem, this doesn't seem like the appropriate fix. The test case shows the randomInterval is doing what it's suppose to do. You can add a print statement to see the values if you want: 

https://github.com/parse-community/Parse-Swift/blob/e3f51f4d5f1520d694edac248ebb883f2c05d375/Tests/ParseSwiftTests/ParseLiveQueryTests.swift#L506-L518

cbaker6: If zeros are the problem, you can make the following change:

```swift
let min = NSDecimalNumber(decimal: Swift.min(30, pow(2, attempts)))
return Int.random(in: 1 ..< Int(truncating: min))
```

cbaker6:
@lsmilek1 thanks for submitting your first PR! Don't forget to add a change log entry: https://github.com/parse-community/Parse-Swift/blob/main/CHANGELOG.md

ParseSwift.playground/Pages/1 - Your first Object.xcplaygroundpage/Contents.swift
cbaker6: Change:

> User `query.first()` of `query.find()`

To: “ Use `query.first()` or `query.find()`”

ParseSwift.playground/Pages/15 - Custom ObjectId.xcplaygroundpage/Contents.swift
```diff
@@ -18,6 +18,11 @@ npm start -- --appId applicationId --clientKey clientKey --masterKey masterKey -
 /*: In Xcode, make sure you are building the "ParseSwift (macOS)" framework.
  */
 
+/*: Warning: A mixed environment with both custom and server generated ObjectId
```
cbaker6, line 21: Change “ObjectId” to “`objectId`“

Sources/ParseSwift/LiveQuery/ParseLiveQuery.swift
lsmilek1: Sorry for misunderstanding, the 0 was my focus and I assumed that without the random Int it is still increasing backoff, so that was the reason for that change. But I tested your update [#209](https://github.com/parse-community/Parse-Swift/pull/209/commits/3d5729921c0376473ee4ea4a37ae04b2e506d074) and I can confirm that it solved even better the reconnection issue that should this commit resolve.

lsmilek1:
Closing this PR as there the LiveQuery reconnection is solved with #209. Thank you!

PR #212, opened 2021-08-06
feat: Add async/await wrappers

cbaker6:

### Issue Description

Add all necessary async/await wrappers. Requires Xcode 13 beta and Swift 5.5.

Developers wanting to use async/await should point their SPM to this branch.

Related issue: #234 

### Approach

Add async/await using wrappers around async completion methods.

### TODOs before merging

- [x] add async/await for all `ParseObject` variants and other types
- [x] add async/await for `ParseAuthentication` types
- [x] bump CI to Xcode 13 
- [x] update documentation for async/await
- [x] add test cases
- [x] add change log entry
- [x] ~~wait for codecov.io to support Xcode 13 codecov~~ Letting codecov slip for now since codecov.io can't read Xcode 13 properly. codecov is really at 85%, will rerun when the website can read latest Xcode.


novemTeam:
@cbaker6 Are `SubscriptionCallback`'s supported? .handleEvent seems to look the same?

cbaker6:
> @cbaker6 Are `SubscriptionCallback`'s supported? .handleEvent seems to look the same?

What would be the use case for using async/await with a subscription callback or with subscriptions for LiveQuery in general?

novemTeam:
> > @cbaker6 Are `SubscriptionCallback`'s supported? .handleEvent seems to look the same?
> 
> What would be the use case for using async/await with a subscription callback or with subscriptions for LiveQuery in general?

I notice that I can't run my own async code this way. Is there an alternative approach I am missing?
I call the function like this
```
let query = NVMFile.query("project" == app.rawValue)
let subscription = query.subscribeCallback!
await handleSynchronisation(subscription: subscription)
```

Schermafbeelding 2021-09-30 om 17 21 16

novemTeam:
@cbaker6 Also another question, what is the correct/best way to compare errors when using async?

I used to check it like this:
```
let query = NVMFile.query()
query.first() { results in
    switch results {
    case .success(_):
        //Do stuff
    case .failure(let error):
        if error.code == ParseError.Code.objectNotFound {
            //Do stuff
        } else {
            //Do other stuff
        }
    }
}
```

i thought of something like this, but I doubt it will be good:

```
do {
    //Stuff
} catch let queryError {
    if let parseError = queryError as? ParseError, parseError.code == ParseError.Code.objectNotFound {
        //Do stuff
    }
    else {
        //Do other stuff
    }
}
```

cbaker6:
I suggest reading https://www.avanderlee.com/swift/async-await/#async-methods-call-in-a-function-that-does-not-support-concurrency

The rest of that article discusses when and how to use async/await and it's important to note it's not the solution for everything.

novemTeam:
> @cbaker6 Also another question, what is the correct/best way to compare errors when using async?
> 
> 
> 
> I used to check it like this:
> 
> ```
> 
> let query = NVMFile.query()
> 
> query.first() { results in
> 
>     switch results {
> 
>     case .success(_):
> 
>         //Do stuff
> 
>     case .failure(let error):
> 
>         if error.code == ParseError.Code.objectNotFound {
> 
>             //Do stuff
> 
>         } else {
> 
>             //Do other stuff
> 
>         }
> 
>     }
> 
> }
> 
> ```
> 
> 
> 
> i thought of something like this, but I doubt it will be good:
> 
> 
> 
> ```
> 
> do {
> 
>     //Stuff
> 
> } catch let queryError {
> 
>     if let parseError = queryError as? ParseError, parseError.code == ParseError.Code.objectNotFound {
> 
>         //Do stuff
> 
>     }
> 
>     else {
> 
>         //Do other stuff
> 
>     }
> 
> }
> 
> ```

@cbaker6 

Do you have an idea on how to cleanly compare `Errors` catched from a `trycatch` with ParseError?

You think my method is the best, or are there better options?

cbaker6:
Your process is fine, though you can shrink the code some:

```swift
do {

    //try Stuff

} catch {

   guard let parseError = error as? ParseError else {
        //Do stuff for different error
      return
    }

    // Handle ParseError.
   switch parseError.code {
                case .objectNotFound: 
                    // Handle object not found
                default:
                    // There was a different issue that we don't know how to handle
                }

}
```

You can see similar in my SDK, though it's currently not using async, https://github.com/netreconlab/ParseCareKit/blob/949393ea1e74eb01ef99c165e1966ff4490a0bae/Sources/ParseCareKit/Objects/Patient.swift#L114-L144

novemTeam:
@cbaker6 Question, would it be possible to merge `1.10.1` into this branch? 

My SPM is set on your async branch, and this way I could test te changes in `1.10.1`

If this isn't an option, no problem, but I think it's worth the question.

Thanks in advance!

PR #222, opened 2021-08-31
Allow mixed custom objectId environment

cbaker6:
Properly allow a mixed custom objectId environment without compromising safety checks using `.save()`. If a developer wants to ignore the objectId checks, they need to specify `isIgnoreCustomObjectIdConfig = true` each time.

Added the following warning to the documentation:

> If you are using `ParseConfiguration.allowCustomObjectId = true` and plan to generate all of your `objectId`'s on the client-side then you should leave `isIgnoreCustomObjectIdConfig = false`. Setting `ParseConfiguration.allowCustomObjectId = true` and `isIgnoreCustomObjectIdConfig = true` means the client will generate `objectId`'s and the server will generate an `objectId` only when the client does not provide one. This can increase the probability of colliding `objectId`'s as the client and server `objectId`'s may be generated using different algorithms. This can also lead to overwriting of `ParseObject`'s by accident as the client-side checks are disabled. Developers are responsible for handling such cases.`

- [x] Add `isIgnoreCustomObjectIdConfig` to saves
- [x] Add test cases
- [x] Add documentation  
- [x] Update change log
- [x] Prepare for release

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,7 +4,7 @@ coverage:
   status:
     patch:
       default:
-        target: 49
+        target: auto
     changes: false
     project:
       default:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,14 @@
 # Parse-Swift Changelog
 
 ### main
-[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.9.6...main)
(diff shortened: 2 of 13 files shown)


cbaker6:
@TomWFox can you look through the documentation when you get a chance?

TomWFox:
I've been on holiday hence the delay ;)

It looks good 👍🏻

Issue #223, opened 2021-08-31
Multiple apps are running the same ParseUser


novemTeam:
### Issue Description
I have an app that uses the ParseSwift SDK. I started a new app a week ago, and I noticed some issues with my other app. I started to look into it and I noticed that my first app had the User credentials of my second app.

This shouldn't happen in any case cause my apps have different Bundle Identifiers. They are not related in any case.

I also noticed that when I log in to one app, it asks me for my Keychain password on the other app. 

I do not have this issue on iOS

### Steps to reproduce
Create 2 projects with different Bundle Identifiers. Add the ParseSwift SDK

###  Environment

Parse Swift
- SDK version: `1.9.4`
- Operating system (iOS, macOS, watchOS, etc.): `macOS`
- Operating system version: `Monterey 12.0 Beta 6`

Server
- Parse Server version: `4.4.0`
- Operating system: `Back4App`

cbaker6:
This is definitely possible because of the way the Keychain is setup. I will take a look at it soon.

novemTeam:
@cbaker6  I don't exactly know how **Parse** handles stuff when it comes to the **Keychain**, but could it be an option to add a part of the **Parse** `applicationId` to the **Keychain** name which stores the _Parse User data_?

cbaker6:
This [post]() states why the issue you mentioned is happening:

> macOS doesn’t have the concept of “local app” keychain [1]. Rather, each app has its own keychain search list and sees all the items in all the keychains in that search list. This search list typically includes the login keychain and the System keychain, but there can be other keychains inserted in the list, either by the user or programmatically. Such keychains can either be filed-based or backed by security token (like a smart card).
>Credentials within the keychain are governed by an access control list, so when you try to use the credential you may end up triggering a user authentication dialog.

Not sure of a solution yet…

novemTeam:
I am currently doing it like this:

```
if error.code == ParseError.Code.invalidSessionToken {
     try? User.logout()
     NotificationCenter.default.post(name: Notification.flUserChanged, object: nil, userInfo: [:])
}
```

But this doesn't really fix the issue. It now just allows the user to login again. Not optimal at all.

cbaker6:
For macOS, might be able to put the whole keychain in a key/value dictionary with the key being the app bundleID, which might solve part of the problem

PR #224, opened 2021-09-04
Use separate Keychain for each app

cbaker6:
Fixes an issue on macOS based ParseSwift apps where they end up using the same Keychain. This is fixed by using the app bundle id for the keychain of each app. The issue is explained here: https://github.com/parse-community/Parse-Swift/issues/223#issuecomment-909711808

**Warning for macOS app developers only** the user who logged in last to your app will have their Keychain upgraded to the patched version. Other users/apps will either need to login again or logout then login again

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,14 @@
 # Parse-Swift Changelog
 
 ### main
-[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.9.7...main)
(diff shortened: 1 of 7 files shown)


cbaker6:
@novemTeam can you try this branch out?

novemTeam:
@cbaker6 I tested it out, I logged out on both the apps, then logged back in. The problem is fixed.
I also tried running the app on iOS back again, just to be sure, and had no issues.

Seems like the issue is fixed with this branch.

Issue #220, opened 2021-08-29
Allow mixed environment of server generated and custom objectId


lsmilek1:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).

### Current Limitation

Currently it is not possible to have mixed environment of custom and server generated objectId. In the case where `User` and user's `Profile` are two different objects in the database (keeping public data only in Profile) there is a need of reference field between these two related objects (Pointer or objectId String).

This is making it complicated in certain scenarios. For example some third object, let it be `Group` has a list of participants `[objectId]` or Pointers. When the group gets deleted and the cloud code function would like to send a message to each remaining participant, it needs to fetch each `Profile`'s data to get related `User`'s objectId to be able to set the ACL of the "group deleted" message object only for that recipient.

### Feature / Enhancement Description

With `allowCustomObjectId = true` allow mixed environment so that it is possible to set the same objectId for multiple objects in different classes. This allows `User` and `Profile` to have the same objectId and there will be no need to fetch additional objects to set the ACL properly as mentioned in the above scenario, reducing also the size of the objects in the database and need of additional indexes on reference field

### Example Use Case

For simple test I removed the `if` check on `saveCommand()` and it works as intended, having the need of not-nil `createdAt` if the object should be updated. 

https://github.com/lsmilek1/Parse-Swift/commit/b3346df5d09dfabeeb9c49e682fb6b8e50fea2bc

Leaving `createdAt = nil` and setting custom objectId to already existing value behaves as intended:
> error saving testInbox: ParseError code=137 error=A duplicate value for a field with unique values was provided

### Alternatives / Workarounds

As mentioned, available workaround is adding extra fields and indexes to the objects and therefore additional load on the server at big scale. In example of `User` and `Profile` it is also easier to read that certain object belongs together

lsmilek1:
Do I miss any security or other reason of that `if ParseSwift.configuration.allowCustomObjectId && objectId == nil { ... }` checks? Because simply leaving them out would allow the behaviour that would noticeably simplify the schema I am using.

cbaker6:
I don’t fully understand what’s being proposed. It seems there are multiple topics mentioned.

[...]

PR #225, opened 2021-09-06
Create witch custom id

lsmilek1:
Added createWithCustomObjectId parameter in .save() functions to remove the need of setting isIgnoreCustomObjectIdConfig = true and allowCustomObjectId = true for a mixed custom objectId environment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,23 @@
 # Parse-Swift Changelog
 
 ### main
-[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.9.7...main)
+[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.9.8...main)
 * _Contributing to this repo? Add info about your change here to be included in the next release_
+
+### 1.9.8
+[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.9.7...1.9.8)
+
+__Fixes__
+- Use a seperate Keychain for each app bundleId. This only effects macOS apps as their Keychain is handled by the OS differently. For macOS app developers only, the user who logged in last to your app will have their Keychain upgraded to the patched version. Other users/apps will either need to login again or logout then login again ([#224](https://github.com/parse-community/Parse-Swift/pull/224)), thanks to [Corey Baker](https://github.com/cbaker6).
+
+__Improvements__
+- Added createWithCustomObjectId parameter in .save() functions to remove the need of setting isIgnoreCustomObjectIdConfig = true and allowCustomObjectId = true for a mixed custom objectId environment. ([#225](https://github.com/parse-community/Parse-Swift/pull/225)), thanks to [Lukas Smilek](https://github.com/lsmilek1)
 
 ### 1.9.7
 [Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.9.6...1.9.7)
 
 __Improvements__
 - Properly allow a mixed custom objectId environment without compromising safety checks using .save(). If a developer wants to ignore the objectId checks, they need to specify isIgnoreCustomObjectIdConfig = true each time ([#222](https://github.com/parse-community/Parse-Swift/pull/222)), thanks to [Corey Baker](https://github.com/cbaker6).
-- Added createWithCustomObjectId parameter in .save() functions to remove the need of setting isIgnoreCustomObjectIdConfig = true and allowCustomObjectId = true for a mixed custom objectId environment
 
 ### 1.9.6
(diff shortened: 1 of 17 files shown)


lsmilek1:
As this is my first time I encountered XCTests and also started updating other branch and per your [comment](https://github.com/parse-community/Parse-Swift/issues/220#issuecomment-911668908) added new branch later, I am really not sure If I got the things right. I apologise in advance!

cbaker6:
Can you resolve the conflicts?

cbaker6:
Also, it looks like you started from the branch I created as opposed to branching from the main

lsmilek1:
Sure, I will have a look on the conflicts. I am sorry, I was never using git control before. I started from non-main branch, you are right (not knowing there will be such conflicts) and after you mentioned it I had actually all the code modified, so I then tried to merge it in the forked main and then create PR... Obviously that is not the way to do that.

cbaker6:
> Also, I did not know how to run the XCTests, as I never did any before and I do not have local Parse Server running (do I need that?) so if you could give me a hint that would be great!

You don't need a parse server to run the tests. You can run them locally using Xcode, https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/testing_with_xcode/chapters/05-running_tests.html or they will run automatically in the CI when you fix the conflicts.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 __Improvements__
 - Properly allow a mixed custom objectId environment without compromising safety checks using .save(). If a developer wants to ignore the objectId checks, they need to specify isIgnoreCustomObjectIdConfig = true each time ([#222](https://github.com/parse-community/Parse-Swift/pull/222)), thanks to [Corey Baker](https://github.com/cbaker6).
+- Added createWithCustomObjectId parameter in .save() functions to remove the need of setting isIgnoreCustomObjectIdConfig = true and allowCustomObjectId = true for a mixed custom objectId environment
```
cbaker6, line 12: Be sure to follow the format of the rest of the change log, you are missing the PR link and contributor information

cbaker6 (changes requested):

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 __Fixes__
 - Use a seperate Keychain for each app bundleId. This only effects macOS apps as their Keychain is handled by the OS differently. For macOS app developers only, the user who logged in last to your app will have their Keychain upgraded to the patched version. Other users/apps will either need to login again or logout then login again ([#224](https://github.com/parse-community/Parse-Swift/pull/224)), thanks to [Corey Baker](https://github.com/cbaker6).
 
+__Improvements__
```
cbaker6, line 13: This is in the wrong place. It should be above 1.9.8 as that version is already released

cbaker6 (changes requested):

cbaker6:
You need to install SwiftLint and ensure you fix all of the errors and warnings before committing or else your builds won't pass. If SwiftLint is installed, you will see the build errors and warnings on your local computer https://github.com/parse-community/Parse-Swift/blob/main/CONTRIBUTING.md

lsmilek1:
I believe this is at the end way over my competence. I already see I am breaking it more than contributing, especially the tests, and in the Swift playgrounds I will have to spin up a local version of the server as I understood your comment. This is all new to me and even I somehow make the tests to pass, the test logic will be rather wrong. 

The goal was to remove the need of `allowCustomObjectId` and `isIgnoreCustomObjectIdConfig` and invert the logic into one optional flag `createWithCustomId` to reduce the code length on each .save() as basically only few objects would make use of the `allowCustomObjectId` + `isIgnoreCustomObjectIdConfig`. I believe a wrapper `.save()` in the non-SDK classes will be more reasonable way to do that instead of attempting to change SDK's essential behavior.

cbaker6:
> I already see I am breaking it more than contributing, especially the tests,

the tests are there to ensure core logic isn’t broken when changes are made. The changes you were making will definitely break core logic, and some of the tests that fail will need to be changed. It also allows us to see what breaks and if the change should be added.

I don’t see a commit where you fixed the linking so we can at least see the test failures to evaluate your changes.

> and in the Swift playgrounds I will have to spin up a local version of the server as I understood your comment.

this is as simple as downloading docker, forking a repo and running a one liner (see my comment below from the readme). If back4app allows you to spin up a free instance you can use that or any other service and tear it down when you are finished.

> This is all new to me and even I somehow make the tests to pass, the test logic will be rather wrong.

I think it’s useful to think about and develop tests for added features and see how/why a change breaks current tests. It will also allow you to learn more about how the SDK is designed and then help you discover bugs in the future

Issue #235, opened 2021-09-14
ParseInstallation not being saved in Keychain


lsmilek1:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Swift SDK](https://github.com/parse-community/Parse-Swift/releases). 

### Issue Description

ParseInstallation.current found not nil only after a fresh install (cleared keychain) and on further app runs found always nil.

### Steps to reproduce

0) clear simulator or device keychain
1) run ParseSwift initialisation in `didFinishLaunchingWithOptions` and print `ParseInstallation.current`
```swift
   func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
        ParseSwift.initialize(applicationId: "...",
                              clientKey: "...",
                              serverURL: URL(string: "https://parseapi.back4app.com")!,
                              liveQueryServerURL: URL(string: "https://felse.b4a.io")!,
                              allowCustomObjectId: true)
        print("installation is nil: \(Installation.current == nil)")
        return true
    }
```
2) restart app without any activity and observe the `ParseInstallation.current` being nil on second and further launch

### Actual Outcome

A `currentInstallationContainer: CurrentInstallationContainer<Self>` is retrieved from Keychain with set properly `installationId` but  with nil `currentInstallation`
>CurrentInstallationContainer<BaseParseInstallation>(currentInstallation: nil, installationId: Optional("f4ec6107-8d0b-4718-801b-33e97c64c3ea"))

### Expected Outcome

`currentInstallation` should have been saved during the first run

### Environment

Client
- Parse Swift SDK version: 1.9.9
- Xcode version: 12.5
- Operating system (iOS, macOS, watchOS, etc.): iOS
- Operating system version: 14.5.

Server
- Parse Server version: 4.5.0
- Operating system: unknown (back4app service provider)
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): unknown (back4app service provider)

Database
[...]

PR #236, opened 2021-09-14
fix: Installation not saving to keychain correctly, returning nil instead

cbaker6:

### Issue Description

Installation not saving to the Keychain correctly, preventing the use of push notifications.

Related issue: #235

### Approach

Save a `BaseParseInstallation` when initializing the SDK that can later be casted to the developers `ParseInstallation`.

When upgrading from previous versions of the SDK, this patch will attempt to fetch the installation from the Parse Server if it's available or else it will create a new one. If you added custom properties to your Installation you should call `Installation.current?.fetch` during SDK initialization to get the latest from the parse-server. Developers should also use cloud code/jobs to clean up any old installations that aren't being used anymore on the server.

### TODOs before merging

- [x] Add tests
- [x] Add entry to changelog
- [x] Add changes to read me
- [x] Remove code duplication in spelling of variables 
- [x] Prepare for new release

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,7 +4,7 @@ coverage:
   status:
     patch:
       default:
-        target: 82
+        target: auto
     changes: false
     project:
       default:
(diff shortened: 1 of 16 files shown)


cbaker6:
@lsmilek1 can you try out this branch and let me know if it works?

lsmilek1:
The problem was solved, now I get installation always. Thank you!

Issue #237, opened 2021-09-20
Better error logging for Invalid Struct


dblythy:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).

### Current Limitation

At the moment, if you define a struct which doesn't match your server payload, you get no meaningful errors that help you isolate the problem.

### Feature / Enhancement Description

Add a `.invalidStruct` error with information as to where the struct doesn't match the server payload

### Example Use Case

-

### Alternatives / Workarounds

-

### 3rd Party References

-

cbaker6:
@dblythy thanks for opening. Can you provide and example of an “invalid struct”? I’ve never created or seen an invalid struct before while using the SDK so I’m interested to see what it looks like. Also, can you post the error or information the SDK currently provides before your PR.

dblythy:
Yes, so it's if your struct doesn't match up with your Parse data that is returned in a query. This happens if the data is already stored using an alternate method, such as the JS SDK, and you're building the `struct` to match the existing data (such as invalid type or not set to optional). At the moment the error is:

```
"Error decoding parse-server response: Optional(<NSHTTPURLResponse: 0x6000019d1100> { URL: http://localhost:1337/1/classes/GameScore } {
```

This changes the error to:

```
"Invalid struct: No value associated with key CodingKeys(stringValue: \"score\", intValue: nil) (\"score\")."
```

cbaker6:
It seems that you can add your comment https://github.com/parse-community/Parse-Swift/issues/237#issuecomment-922748352 in the issue description for "Example Use Case" and "Alternative Workarounds"

PR #238, opened 2021-09-20
Add invalid struct error

dblythy:

### Issue Description

Invalid structs are really hard to isolate the cause of the problem, especially with larger classes

Related issue: #237

### Approach

Checks if error code is `4865` and returns error .invalidStruct if so

### TODOs before merging

- [x] Add tests
- [ ] Add entry to changelog

### DIFF
--- a/Sources/ParseSwift/API/URLSession+extensions.swift
+++ b/Sources/ParseSwift/API/URLSession+extensions.swift
@@ -66,6 +66,10 @@ extension URLSession {
                           let json = try? JSONSerialization
                             .data(withJSONObject: responseData,
                               options: .prettyPrinted) else {
+                        let err = (error as NSError)
+                        if err.code == 4865, let description = err.userInfo["NSDebugDescription"] {
+                            return .failure(ParseError(code: .invalidStruct, message: "Invalid struct: \(description)"))
+                        }
                         return .failure(ParseError(code: .unknownError,
                                                    // swiftlint:disable:next line_length
                                                    message: "Error decoding parse-server response: \(response) with error: \(error.localizedDescription) Format: \(String(describing: String(data: responseData, encoding: .utf8)))"))
--- a/Sources/ParseSwift/Types/ParseError.swift
+++ b/Sources/ParseSwift/Types/ParseError.swift
@@ -351,6 +351,11 @@ public struct ParseError: ParseType, Decodable, Swift.Error {
          a non-2XX status code.
          */
         case xDomainRequest = 602
+        
+        /**
+         Error code indicating the Parse Swift struct is invalid..
+         */
+        case invalidStruct = 603
 
         /**
          Error code indicating any other custom error sent from the Parse Server.
--- a/Tests/ParseSwiftTests/ParseQueryTests.swift
+++ b/Tests/ParseSwiftTests/ParseQueryTests.swift
@@ -723,6 +723,17 @@ class ParseQueryTests: XCTestCase { // swiftlint:disable:this type_body_length
         }
 
         let query = GameScore.query()
+        do {
+            _ = try query.first(options: [])
+            XCTFail("Should have thrown error")
+        } catch {
+            guard let error = error as? ParseError else {
+                XCTFail("Should have casted as ParseError")
+                return
+            }
+            XCTAssertEqual(error.message, "Invalid struct: No value associated with key CodingKeys(stringValue: \"score\", intValue: nil) (\"score\").")
+            XCTAssertEqual(error.code, .invalidStruct)
+        }
         XCTAssertThrowsError(try query.first(options: []))
     }
 


Sources/ParseSwift/API/URLSession+extensions.swift
```diff
@@ -66,6 +66,10 @@ extension URLSession {
                           let json = try? JSONSerialization
                             .data(withJSONObject: responseData,
                               options: .prettyPrinted) else {
+                        let err = (error as NSError)
```
cbaker6, line 69: Shorthand variable names like `err` are not the swift way and we don't use this type of naming in the SDK. You should make the name change along with other changes:

```swift
let nsError = error as NSError
if nsError.code == 4865,
  let description = nsError.userInfo["NSDebugDescription"] {
   return .failure(ParseError(code: .invalidStruct, message: "Invalid struct: \(description)"))
}
```

Sources/ParseSwift/Types/ParseError.swift
```diff
@@ -352,6 +352,11 @@ public struct ParseError: ParseType, Decodable, Swift.Error {
          */
         case xDomainRequest = 602
 
+        /**
+         Error code indicating the Parse Swift struct is invalid.
+         */
+        case invalidStruct = 603
```
cbaker6, line 358: I don't see `603` as a code in the JS SDK or server. What happens if another error code is put in it's place at the server and JS SDK level? It seems this number should be something stable or internal to the SDK. Unless you plan to introduce this error at the JS and server level?

Sources/ParseSwift/Types/QueryViewModel.swift
```diff
@@ -38,6 +38,7 @@ open class QueryViewModel<T: ParseObject>: QueryObservable {
     }
 
     /// Updates and notifies when there is an error retrieving the results.
+    // swiftlint:disable:next redundant_optional_initialization
```
cbaker6, line 41: You can remove this line

Sources/ParseSwift/Types/QueryViewModel.swift
```diff
@@ -38,6 +38,7 @@ open class QueryViewModel<T: ParseObject>: QueryObservable {
...
     open var error: ParseError? = nil {
```
cbaker6, line 42: change this to:

```swift
open var error: ParseError? {
```

Sources/ParseSwift/Types/CloudViewModel.swift
```diff
@@ -29,6 +29,7 @@ open class CloudViewModel<T: ParseCloud>: CloudObservable {
     }
 
     /// Updates and notifies when there is an error retrieving the results.
+    // swiftlint:disable:next redundant_optional_initialization
```
cbaker6, line 32: Remove this line

Sources/ParseSwift/Types/CloudViewModel.swift
```diff
@@ -29,6 +29,7 @@ open class CloudViewModel<T: ParseCloud>: CloudObservable {
...
     open var error: ParseError? = nil {
```
cbaker6, line 33: Change this to:

```swift
open var error: ParseError? {
```

cbaker6 (changes requested):

cbaker6:
It also looks like you broke some Linux tests as well. Those will need to be fixed before the PR can be approved.

Sources/ParseSwift/Types/ParseError.swift
cbaker6: Changing to `case invalidStruct` doesn't resolve the problem. If you are not adding this error to server and JS SDK, then the liens added to ParseError should be removed and you should be using `.unknownError` as it's currently the way internal SDKs errors are handled. You current solution can cause a major issue when a new error is introduced at the server level as it will decode improperly when using the Swift SDK

dblythy:
The test that is failing on Linux is the one that I added (the different response for the .invalidStruct error). Does linux handle swift json decoding differently?

Sources/ParseSwift/Types/ParseError.swift
dblythy: Ah right, I understand!

Tests/ParseSwiftTests/ParseQueryTests.swift
```diff
@@ -723,6 +723,18 @@ class ParseQueryTests: XCTestCase { // swiftlint:disable:this type_body_length
...
+        do {
+            _ = try query.first(options: [])
+            XCTFail("Should have thrown error")
+        } catch {
+            guard let error = error as? ParseError else {
+                XCTFail("Should have casted as ParseError")
+                return
+            }
+            // swiftlint:disable:next line_length
+            XCTAssertEqual(error.message, "Invalid struct: No value associated with key CodingKeys(stringValue: \"score\", intValue: nil) (\"score\").")
```
cbaker6, line 735: Try replacing these lines with:

```swift
#if !os(Linux) && !os(Android)
            // swiftlint:disable:next line_length
            XCTAssertEqual(error.message, "Invalid struct: No value associated with key CodingKeys(stringValue: \"score\", intValue: nil) (\"score\").")
            #endif
```

cbaker6 (changes requested):

cbaker6:
> The test that is failing on Linux is the one that I added (the different response for the .invalidStruct error). Does linux handle swift json decoding differently?

Yes there are some differences with key order and it's possible linux throws a different NSError as the error code you are using might come from obj-c. If you want to see see the error code, you probably will need to add a print statement and then check the linux CI for the value

cbaker6:
Let me know if you need any help with finishing this. After this PR is merged, I'll release a new version.

dblythy:
Sorry I’ll try wrap this up today!

novemTeam:
@dblythy Will this also show better error logging for when a certain key that is required in the database isn't filled in?

dblythy:
Still testing locally to make sure this doesn't clash with other error codes

dblythy:
@novemTeam I'm not sure to be honest, what error do you currently get when that happens?

Issue #239, opened 2021-09-22
Verification Emails are triggered every time User.current is saved


dblythy:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Swift SDK](https://github.com/parse-community/Parse-Swift/releases). 

### Issue Description

Parse Swift triggers email verification whenever `User.current?.save` is called.

### Steps to reproduce

1. Turn `verifyUserEmails` on.
2. Update any field on `User.current`
3. Check emails

### Actual Outcome

Emails are sent every time user is updated

### Expected Outcome

Emails to only send on signup, or email update, as per the SDK.

Specifically, the issue seems to be related to here o the server:

```js
RestWrite.prototype._validateEmail = function () {
 if (!this.data.email || this.data.email.__op === 'Delete') {
    return Promise.resolve();
  } 
```

For the JS SDK, `this.data.email` is undefined on any update. For Parse Swift calls, this.data.email is passed.

### Environment

Client
- Parse Swift SDK version: `1.9.10`
- Xcode version: `10.15.7`
- Operating system (iOS, macOS, watchOS, etc.): `iOS`
- Operating system version: `14.3`

Server
- Parse Server version: `Master (around 4.5.0`
- Operating system: `macOS`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `local`

Database
- System (MongoDB or Postgres): `mongoDB`
- Database version: `4.4`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `Atlas`

cbaker6:
In the JS SDK, are you able to show where the following occurs:

> For the JS SDK, this.data.email is undefined on any update

That will provide some guidance on where this is happening on the client level, or if the server is actually handling it.

dblythy:
[...]

PR #241, opened 2021-09-22
fix: Don't encode unmodified email when updating a user

cbaker6:

### Issue Description

When a current user email isn't modified, it shouldn't be sent to the server or else it will trigger verify email.

Related issue: #239 

### Approach

Only encode email if it's modified.

### TODOs before merging

- [x] Add tests
- [x] Add entry to changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 [Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.9.10...main)
 * _Contributing to this repo? Add info about your change here to be included in the next release_
 
+__Fixes__
+- ParseUser shouldn't send email if it hasn't been modified or else email verification is resent ([#241](https://github.com/parse-community/Parse-Swift/pull/241)), thanks to [Corey Baker](https://github.com/cbaker6).
+
 ### 1.9.10
 [Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.9.9...1.9.10)
 __Fixes__
--- a/Sources/ParseSwift/Objects/ParseUser.swift
+++ b/Sources/ParseSwift/Objects/ParseUser.swift
@@ -20,7 +20,7 @@ public protocol ParseUser: ParseObject {
     Determines if the email is verified for the `ParseUser`.
      - note: This value can only be changed on the Parse Server.
     */
-    var emailVerified: Bool? { get set }
+    var emailVerified: Bool? { get }
 
     /**
      The password for the `ParseUser`.
@@ -945,7 +945,11 @@ extension ParseUser {
 
     private func updateCommand() -> API.Command<Self, Self> {
         var mutableSelf = self
-        mutableSelf.emailVerified = nil
+        if let currentUser = Self.current,
+           currentUser.hasSameObjectId(as: mutableSelf) == true,
+           currentUser.email == mutableSelf.email {
+            mutableSelf.email = nil
+        }
         let mapper = { (data) -> Self in
             try ParseCoding.jsonDecoder().decode(UpdateResponse.self, from: data).apply(to: self)
         }
--- a/Sources/ParseSwift/Objects/ParseUser.swift
+++ b/Sources/ParseSwift/Objects/ParseUser.swift
@@ -20,7 +20,7 @@ public protocol ParseUser: ParseObject {
     Determines if the email is verified for the `ParseUser`.
      - note: This value can only be changed on the Parse Server.
     */
-    var emailVerified: Bool? { get }
+    var emailVerified: Bool? { get set }
 
     /**
      The password for the `ParseUser`.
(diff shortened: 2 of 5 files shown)


cbaker6:
@dblythy can you try out this branch and see if it fixes your problem?

If it works, it seems the server is allowing `emailVerified` to be changed from the client, which it shouldn't which is causing the email verification to be triggered.

dblythy:
I sure can, give me a moment

dblythy:
See comment here https://github.com/parse-community/Parse-Swift/issues/239#issuecomment-924818217

cbaker6:
@dblythy Try this, if it works, I'll add the necessary testcases.

dblythy:
I will have a look tomorrow morning if that's ok. 

So there's no way to only send mutated keys to the server using Parse Swift? I'm worried this will cause race conditions with int fields, and unnecessary database ops. 

Alternatively I could use `.revert()` in cloud functions to prevent the database ops.

dblythy:
Just a quick comment - whilst this will likely fix the issue, the core problem of `dirtyKeys` still remains. This is problematic because:

- unnecessary data is sent to parse server, which in developing countries may result in unbearable latency, or, on a high scaled system, result in additional data expenses 
- unnecessary database ops will happen when no data is actually changing 
- Cloud functions will tell developers to assume keys are dirty when they are not
- email being set to dirty is just an example of an unpredictable consequences of all the keys being sent, there could be other cases which we don’t know about. 

@cbaker6 is there a way we could perhaps store the serverData from the initial fetch of the object, and then compare changes to keys onSave if we can’t directly listen to property changes on a struct? I’m not overly familiar with swift but happy to look into creative solutions if it helps.

cbaker6:
> So there's no way to only send mutated keys to the server using Parse Swift?

There are a few ways, but the developer has to take some action here. For example, developers can add the following method (or a similar one) to their `ParseObject`'s:

```swift
struct MyUser: ParseUser {
    var username: String?
    var email: String?
    var emailVerified: Bool?
    var password: String?
    var objectId: String?
    var createdAt: Date?
    var updatedAt: Date?
    var ACL: ParseACL?

    // method that creates empty versions of self
    func emptyObject() -> Self {
        var object = Self()
        object.objectId = objectId
        return object
    }
}
```

Then add properties what will be mutated and saved. Only these get sent to the server:

```swift
var myObject = MyUser()
myObject.email = "myNewEmail@parse.com"
myObject.save { _ in }
```

This also can be easily achieved by using `ParseOperation`: https://github.com/parse-community/Parse-Swift/blob/ddcf35f54365397f75068744e6ff975e875dd73a/Tests/ParseSwiftTests/ParseOperationTests.swift#L100-L103

These examples should address most of your comments below:

> Just a quick comment - whilst this will likely fix the issue, the core problem of `dirtyKeys` still remains. This is problematic because:
> 
> * unnecessary data is sent to parse server, which in developing countries may result in unbearable latency, or, on a high scaled system, result in additional data expenses
> * unnecessary database ops will happen when no data is actually changing
> * Cloud functions will tell developers to assume keys are dirty when they are not
> * email being set to dirty is just an example of an unpredictable consequences of all the keys being sent, there could be other cases which we don’t know about.

When it comes to:
> @cbaker6 is there a way we could perhaps store the serverData from the initial fetch of the object, and then compare changes to keys onSave if we can’t directly listen to property changes on a struct?

"Listening to property changes" is typically associated to `reference types` or "instances of Classes" in many languages, not just Swift. Most/All of the Parse SDK's use "classes" with the exception of the Swift SDK. `value types` or "structs" don't have this ability and are one of the distinct difference between a `struct` and a `class`. The Swift SDK doesn't need to listen property changes and instead should use a method similar to what I mentioned at the beginning of this comment. It's important not to think or compare of the Swift SDK to the other SDKs as it has a completely different design philosophy. Flo's [comments](https://github.com/parse-community/Parse-Swift/issues/3#issuecomment-328351357) provide good insight into the design philosophy and the reasoning. For these reasons, the Swift SDK can exist with zero dependencies, not run into the threading issues the other client SDKs encounter, along with having `ParseObject`'s play nicely with MVVM and SwiftUI (from what I see `reference types` is what's causing a major barrier when using MVVM in the issues you and others mention with the JS SDK).

cbaker6:
> I'm worried this will cause race conditions with int fields

How would ParseSwift cause race conditions where the other SDK's wouldn't?

cbaker6:
if you have questions about my previous comments feel free to continue commenting even though this PR is closed.

dblythy:
Should we move this issue to a discussion regarding `dirty` so other users can find it easily if need be?

cbaker6:
> Should we move this discussion to an issue regarding `dirty` so other users can find it easily if need be?

Sure!

dblythy:
Just confirming that this PR has fixed the problem for me 😊  Sorry it took a while to test, being in Australia means timezones become complicated.

Thanks again @cbaker6!

Issue #242, opened 2021-09-22
Partially updating an object sends the full object to the server


dblythy:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Swift SDK](https://github.com/parse-community/Parse-Swift/releases). 

### Issue Description

When fetching a saved object and then calling `.save`, all keys are marked as `dirty` and saved to the backend, causing unnecessary traffic and database ops. 

### Steps to reproduce

1. Fetch an existing object
2. Have a Parse Cloud beforeSave trigger that logs `object.dirtyKeys()`. 

### Actual Outcome

All fields are dirty. 

### Expected Outcome

No fields should be dirty unless mutated. 

### Environment

Client
- Parse Swift SDK version: `1.9.10`
- Xcode version: `10.15.7`
- Operating system (iOS, macOS, watchOS, etc.): `iOS`
- Operating system version: `14.7`

Server
- Parse Server version: `4.5.0`
- Operating system: `macOS`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `localhost`

Database
- System (MongoDB or Postgres): `mongodb`
- Database version: `4.2`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `Atlas`

cbaker6:
Posting my responses to your comments in #241:

> So there's no way to only send mutated keys to the server using Parse Swift?

There are a few ways, but the developer has to take some action here. For example, developers can add the following method (or a similar one) to their `ParseObject`'s:

```swift
struct MyUser: ParseUser {
    var authData: [String: [String: String]?]?
    var username: String?
    var email: String?
    var emailVerified: Bool?
    var password: String?
    var objectId: String?
    var createdAt: Date?
    var updatedAt: Date?
    var ACL: ParseACL?

    // method that creates empty versions of self
    func emptyObject() -> Self {
        var object = Self()
        object.objectId = objectId
        return object
    }
}
```

Then add properties what will be mutated and saved. Only these get sent to the server:

[...]

PR #243, opened 2021-09-23
feat: Add emptyObject property for sending modified keys to server

cbaker6:

### Issue Description

When fetching a saved object and then calling .save, all keys are marked as dirty and saved to the backend, causing unnecessary traffic and database ops.

Related issue: #242

### Approach

Developers were able to accomplish this before this PR by creating a new instance of their `ParseObject`, setting the `objectId` and `createdAt` to the object they want to modify, and then only modifying the needed keys. They could also do any of the following mentioned in https://github.com/parse-community/Parse-Swift/issues/242#issuecomment-925413030.

This PR makes the process easier with less work on the developer by adding a computed property called `emptyObject` to all `ParseObject`s that provides a copy of the respective ParseObject that only contains the current `objectId` and `createdAt`. If developers wants to only modify a subset of the keys, they can modify a mutable version of the empty object. When the mutations of the empty object are saved, only those keys are sent to the server as opposed to sending keys that were not modified before the save. Example's on how to use are in the playgrounds as well as below:

```swift
//: Define initial GameScores.
let score = GameScore(score: 10)

/*: Save asynchronously (preferred way) - Performs work on background
    queue and returns to specified callbackQueue.
    If no callbackQueue is specified it returns to main queue.
*/
score.save { result in
    switch result {
    case .success(let savedScore):

        /*: To modify, need to make it a var as the value type
            was initialized as immutable. Using `emptyObject`
            allows you to only send the updated keys to the
            parse server as opposed to the whole object.
        */
        var changedScore = savedScore.emptyObject
        changedScore.score = 200
        changedScore.save { result in
            switch result {
            case .success:
                /*: Only the modified keys were sent to produce a successful save.
                */

            case .failure(let error):
                assertionFailure("Error saving: \(error)")
            }
        }
    case .failure(let error):
        assertionFailure("Error saving: \(error)")
    }
}
```

A "possible breaking change" is introduced depending on how developers have created initializers for their `ParseObject`s. If a developer had a `ParseObject` with inits that looked like:

```swift
//: Create your own value typed `ParseObject`.
    //: Those are required for Object
    var objectId: String?
    var createdAt: Date?
    var updatedAt: Date?
    var ACL: ParseACL?

    //: Your own properties.
    var score: Int = 0

    init(score: Int) {
        self.score = score
    }

    init(objectId: String?) {
        self.objectId = objectId
    }
}
```

it is recommended to leverage `extensions` to preserve the convenience initializer by converting to the following:

```swift
//: Create your own value typed `ParseObject`.
    //: Those are required for Object
    var objectId: String?
    var createdAt: Date?
    var updatedAt: Date?
    var ACL: ParseACL?

    //: Your own properties.
    var score: Int = 0
}

//: It's recommended to place custom initializers in an extension
//: to preserve the convenience initializer.
extension GameScore {

    init(score: Int) {
        self.score = score
    }

    init(objectId: String?) {
        self.objectId = objectId
    }
}
```

If developers already implemented the empty initializer, `init()` or already use `extensions` for inits, then no breaks to your code should occur. If the developer has required fields (non-optional fields) in their ParseObject's they have should already been specifying values or marking them as optional anyways as this prevents decoding the respective objects as parse `Pointer`'s(see https://github.com/parse-community/Parse-Swift/issues/157#issuecomment-858671025 for more info).

### TODOs before merging

- [x] Add tests
- [x] Add entry to changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,21 @@
 # Parse-Swift Changelog
 
 ### main
-[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.9.10...main)
+[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.10.0...main)
 * _Contributing to this repo? Add info about your change here to be included in the next release_
+
+### 1.10.0
+[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.9.9...1.10.0)
+
+__Improvements__
+- (Breaking Change) Provide ParseObject method emptyObject that makes it easy to send only modified keys to the server. This change "might" be breaking depending on your implementation as it requires ParseObjects to now have an empty initializer, init() ([#243](https://github.com/parse-community/Parse-Swift/pull/243)), thanks to [Corey Baker](https://github.com/cbaker6).
 
 __Fixes__
 - ParseUser shouldn't send email if it hasn't been modified or else email verification is resent ([#241](https://github.com/parse-community/Parse-Swift/pull/241)), thanks to [Corey Baker](https://github.com/cbaker6).
 
 ### 1.9.10
 [Full Changelog](https://github.com/parse-community/Parse-Swift/compare/1.9.9...1.9.10)
+
 __Fixes__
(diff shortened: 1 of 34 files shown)


cbaker6:
@dblythy let me know what you think about the method name, `emptyObject`. If you think of a more applicable name, feel free to suggest.

Issue #247, opened 2021-10-01
Custom init() is not recognized anymore


novemTeam:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Swift SDK](https://github.com/parse-community/Parse-Swift/releases). 

### Issue Description
There is a problem when using custom init() code on a ParseObject class.
I get the error `Type 'User' does not conform to protocol 'ParseObject'` on every ParseObject struct I use.
My code used to work, and all of a sudden it is broken.

<img width="1159" alt="Schermafbeelding 2021-10-01 om 17 59 22" src="https://user-images.githubusercontent.com/44820440/135654079-7eef5f33-5f92-4ae6-a1a9-afbdfe3347d5.png">
<img width="1099" alt="Schermafbeelding 2021-10-01 om 18 00 02" src="https://user-images.githubusercontent.com/44820440/135654088-d94b8051-1cb9-4d77-b9e3-07399e6b44ba.png">

### Steps to reproduce
Add a init() function to your Parse struct, and add variables to it.

### Actual Outcome
Unexpected error

### Expected Outcome
Nothing

### Environment

Client
- Parse Swift SDK version: `1.10.0`
- Xcode version: `13.0`
- Operating system (iOS, macOS, watchOS, etc.): `SPM`
- Operating system version: `12.0`

Server
- Parse Server version: `4.4.0`
- Operating system: `Back4App`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `AWS`

Database
- System (MongoDB or Postgres): `MongoDB`
- Database version: `3.6`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `?`

cbaker6:
This is due to you updating to `1.10.0`, please see the "Breaking Change" notes: https://github.com/parse-community/Parse-Swift/releases/tag/1.10.0 which point to https://github.com/parse-community/Parse-Swift/pull/243#issue-1005536964.

[...]

PR #248, opened 2021-10-02
feat: Add Set operation to allow for single key updates

dblythy:

### Issue Description

Currently, the solution to only updating one field on an object is to use `emptyObject`, which has some side-effects.

Related issue: #242

### Approach

This adds a new simple operation of `.set`, so that individual keys can be updated without requiring a custom init.

Although `.set` is perhaps an artefact of the old SDK, this would allow developers to use `.save` to save the whole object, and `.operation.set` to update the object, which in my opinion, is a good balance.

If we agree on this approach, and I haven't missed any side-effects such as #247, I think we can revert `emptyObject`.

Let me know your thoughts!

### TODOs before merging

- [x] Add tests
- [ ] Add entry to changelog

### DIFF
--- a/Sources/ParseSwift/Operations/ParseOperation.swift
+++ b/Sources/ParseSwift/Operations/ParseOperation.swift
@@ -26,6 +26,18 @@ public struct ParseOperation<T>: Savable where T: ParseObject {
     }
 
     /**
+     An operation that sets a field's value.
+     - Parameters:
+        - key: The key of the object.
+        - value: The value to set it to
+     */
+    public func set<W>(_ key: String, value: W) -> Self where W: Encodable {
+        var mutableOperation = self
+        mutableOperation.operations[key] = value
+        return mutableOperation
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


dblythy:
Adds two new options to operation:

- `.set`, which sends the field if it has been updated since last fetch
- `.forceSet`, which force sets the field, regardless of last fetch (similar to current behaviour on `struct`

cbaker6:
In general, you are attempting to use the style of the other SDK's with JSON dictionaries. The Swift SDK doesn't use this style outside of test cases and encourages future coding behaviors that's outside of the clean coding style and native encoding/decoding of this SDK.

I've changed your code previously when you attempted this in your first PR, https://github.com/parse-community/Parse-Swift/pull/165/commits/5ebb164f9beb7f822cfc7800211894e71e1692d9

dblythy:
Yep, question answered. I just couldn't figure out how to write the test case to emulate the save op, so thanks for that!

No worries, I'll get to the other PR in the next few hours. Thanks again!!

Issue #264, opened 2021-10-21
Setting property to nil doesn't save to database


Vortec4800:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Swift SDK](https://github.com/parse-community/Parse-Swift/releases). 

### Issue Description
Setting a value on an optional model property doesn't propagate to the server.

### Steps to reproduce
```swift
struct TestObject: ParseObject {
	// ParseObject Conformance
	var objectId: String?
	var createdAt: Date?
	var updatedAt: Date?
	var ACL: ParseACL?

	var testValue: String?
}

var testObj = TestObject()
testObj.testValue = "value"

testObj = try! testObj.save()
// Break here and check value on dashboard

testObj.testValue = nil

testObj = try! testObj.save()
// Break here and check value again on dashboard, see that it still says "value" and not (undefined)
```

### Expected Outcome
Final value on dashboard should be (undefined) or nil, instead of keeping the old value.

### Environment

Client
- Parse Swift SDK version: `2.0.2`
- Xcode version: `13.0`
- Operating system (iOS, macOS, watchOS, etc.): `macOS`
- Operating system version: `11.6`

Server
- Parse Server version: `4.10.4`
- Operating system: `Ubuntu`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `Digital Ocean`

Database
- System (MongoDB or Postgres): `MongoDB`
- Database version: `4.4`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `Digital Ocean`

cbaker6:
Closing this due to no activity and the previous answers providing the capability

PR #268, opened 2021-10-28
Setting property to nil causes delete op on save

Vortec4800:

### Issue Description
Currently, saving a `nil` value to a property updates that value to `nil` in the local SDK, but the value on the server does not get changed.

Related issue: #264 

### Approach
This approach was inspired by the way the Javascript SDK and REST API handles unsetting values. Instead of updating a value to a new value, a custom Delete operation is sent instead. This property wrapper looks for a `nil` value and if found, codes the Delete operation as the value.

### TODOs before merging

- [ ] Add tests
- [ ] Add entry to changelog

### DIFF
--- a/ParseSwift.playground/Pages/1 - Your first Object.xcplaygroundpage/Contents.swift
+++ b/ParseSwift.playground/Pages/1 - Your first Object.xcplaygroundpage/Contents.swift
@@ -39,6 +39,9 @@ struct GameScore: ParseObject {
 
     //: Your own properties.
     var score: Int = 0
+	
+	//: Optional custom properties need to be marked with @NullableProperty or setting properties to `nil` won't propagate to server
+	@NullableProperty var gameEndDate: Date?
 
     /*:
      It's recommended the developer adds the emptyObject computed property or similar.
@@ -76,10 +79,10 @@ struct GameData: ParseObject {
     var ACL: ParseACL?
 
     //: Your own properties.
-    var polygon: ParsePolygon?
+	@NullableProperty var polygon: ParsePolygon?
     //: `ParseBytes` needs to be a part of the original schema
     //: or else you will need your masterKey to force an upgrade.
-    var bytes: ParseBytes?
+	@NullableProperty var bytes: ParseBytes?
 }
 
 //: It's recommended to place custom initializers in an extension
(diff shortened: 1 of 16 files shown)


cbaker6:
Just for clarity:

> In the Swift SDK this isn't possible currently. The closest we could get would be:
>
>```swift
>var message...
>message.league = league
>message.sender = requestingUser
>message.messageBody = messageBody
>	
>message = try newLeague.save()
>
>let removeOperation = message.operation.unset("unread")
>message = try removeOperation.save()
>```
>
>Similar but now we're running two separate operations and depending on code layout you may not always know at the >time that save is being called that you might be setting a value to nil. You also have cloud functions getting called twice >- there may be a way to combine those operations into a single transaction but that adds to the complexity even >further.

You can set/unset all within the same `save`:

```swift
var message...
let operations = message.operation
   .set(("league", \. league), league)
   .set(("sender", \.sender), requestingUser)
   .set(("messageBody", \.messageBody), messageBody)
   .unset("unread")
message = try operations.save()
```

Which looks very similar to your android example and won't make extra calls to Cloud Code nor does it add complexity.

cbaker6:
> Specifically, the change is in the case .none action, the standard implementation encodes the nil/null value directly in the JSON which looks like {"key": null} in the encoded output. This is technically accepted by the server, however values are stored as "null" instead of "undefined" in the backend and this will cause some queries to not work as expected. The modified version is specific to Parse and uses the Delete operation to encode that operation as the value of the key. This tells the server to remove the value for that key, same as the operation would, but it happens as part of the larger save instead of in a separate operation.

@Vortec4800 from your comments above, I'm assuming your DB is Mongo as there is no difference between `null` and `undefined` for Postgres. The [Swift SDK 3.0.0](https://github.com/parse-community/Parse-Swift/releases/tag/3.0.0) should address much of your comment above, specifically, #308.

cbaker6:
It seems #315 should help you carry out this PR, particularly checking if `originalData` is available and creating a modified `shouldRestoreKey` to do `shouldSetKeyToNil`.

You will probably want two wrappers that almost do the same thing. 1) encodes null and 2) deletes the item. This will be important for mongo because of what I mentioned here: https://github.com/parse-community/Parse-Swift/pull/268#issuecomment-1005843653. Essentially, developers can choose which wrapper works best for them.

cbaker6:
Closing this due to no activity and the previous answers providing the capability

Issue #226, opened 2021-09-06
Wrong image data when fetching image


vdkdamian:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Swift](https://github.com/parse-community/Parse-Swift/releases). 
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases).

### Issue Description
I am seeing issues when fetching an image. When i fetch an image for the first time, it loads the image corretly. But when it then tries to load for the second time, the image is broken. I looked at what the downloaded image looks like when opening in a text editor. 

I noticed it doesn't contain data, but a path. This is an example:
`file:\/\/\/var\/folders\/y5\/3qbps9t11cb2fz7lh1x8bw7w0000gn\/T\/com.app.myApp\/CFNetworkDownload_rDopeR.tmp`

My uploaded images in the Database are valid images. 

I want to point out that this way of fetching images always worked for me. It's since the Beta's that there is something wrong, but I don't find anything on the Apple Developer known issues page. 

### Steps to reproduce
This is an example of code I use.

```
let modelQuery =  Model.query()
modelQuery.find { result in
    switch result {
    case .success(let model):
        if let imageURL = model.image?.localURL {
            print("imageURL: \(imageURL)")
            if let imgData = try? Data(contentsOf: imageURL) {
                print("imgData: \(imgData)")
                print("imgData as string: \(String(decoding: imgData, as: UTF8.self))")
                self.imageData = imgData
            }
        } else {
            model.image?.fetch(completion: { result in
                switch result {
                case .success(let fetchedImage):
                    if let localImageUrl = fetchedImage.localURL {
                        print("localImageUrl: \(localImageUrl)")
                        if let imgData = try? Data(contentsOf: localImageUrl) {
                            print("imgData: \(imgData)")
                            print("imgData as string: \(String(decoding: imgData, as: UTF8.self))")
                            self.imageData = imgData
                        }
                    }
                case .failure(_):
                    break
                }
            })
        }
        self.model = model
    case .failure(_):
        break
    }
}
```

This is my log:

[...]

PR #272, opened 2021-11-02
fix: don't use cache for fetching objects and files by default

cbaker6:

### Issue Description

When objects or files are being “fetched”, they sometimes leverage the OS level cache and pull from cache if there was a cache hit. This can lead to unintended behavior as when a “fetch” is used, it implies fresh data from the server.

In addition, when caching a file that was downloaded via the SDK, the temporary file location is cached instead the new location the file was moved to. This results in a cache hit, but points to a file location that’s no longer valid.

Related issue: #226 

### Approach

For "fetching", the default cache policy is set for all `ParseObject`'s and `ParseFile`'s to not use cache. If the developer wants to use cache, they should pass the respective cache policy into `options` when calling the respective fetch method. The default cache policy for the SDK can be changed by the developer if needed: https://github.com/parse-community/Parse-Swift/blob/a64726fde852cbfd8f1e2a67fe0d9c3fe4c17872/Sources/ParseSwift/Parse.swift#L33-L37

When fetching `ParseFile`, cache the final location the file was moved to instead of the temporary location

### TODOs before merging

- [x] Add tests
- [x] Add entry to changelog


cbaker6:
@vdkdamian can you try out this branch? Note that it will only fix your problem if was related to the caching I mentioned in https://github.com/parse-community/Parse-Swift/issues/226#issuecomment-913864845 and https://github.com/parse-community/Parse-Swift/issues/226#issuecomment-958012896. If this doesn't fix your issue, then you will have to look in other places in your code and SDK.

vdkdamian:
@cbaker6 This solves my issue! I tested it multiple times on iOS, then multiple times on macOS. Everything is working fine for me now.

PR #294, opened 2021-12-02
feat: improve ParseRelation by adding more methods

cbaker6:

### Issue Description

`ParseRelation` can use some additional methods to make it easier to create and query relations.

Related issue: #n/a

### Approach

Add the following:

- `ParseRelation`: `func add<U>(_ objects: [U]) throws -> ParseOperation<T> where U: ParseObject`
- `ParseRelation`: `func remove<U>(_ objects: [U]) throws -> ParseOperation<T> where U: ParseObject`
- `ParseRelation`: `func query<U>(_ key: String, parent: U) throws -> Query<T> where U: ParseObject`
- `ParseRelation`: `func query<U>(_ key: String, child: U) throws -> Query<U> where U: ParseObject`
- `ParseRelation` Playground save/query example
- `QueryConstraint`: `func related(key: String) -> QueryConstraint`
- `QueryConstraint`: `func related <T>(object: T) throws -> QueryConstraint where T: ParseObject`
- `QueryConstraint`: `func related <T>(object: Pointer<T>) -> QueryConstraint where T: ParseObject`

Removed the following:

- optional targets/parents from `ParseOperation` and `ParseRelation` to get rid of unnecessary unwrapping

### TODOs before merging

- [x] Add tests
- [ ] Add entry to changelog

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,7 +6,7 @@ coverage:
   status:
     patch:
       default:
-        target: auto
+        target: 74
     changes: false
     project:
       default:
--- a/ParseSwift.playground/Pages/12 - Roles and Relations.xcplaygroundpage/Contents.swift
+++ b/ParseSwift.playground/Pages/12 - Roles and Relations.xcplaygroundpage/Contents.swift
@@ -47,6 +47,31 @@ struct Role<RoleUser: ParseUser>: ParseRole {
     }
 }
 
+//: Create your own value typed `ParseObject`.
+struct GameScore: ParseObject, ParseObjectMutable {
+    //: These are required by ParseObject
+    var objectId: String?
+    var createdAt: Date?
+    var updatedAt: Date?
+    var ACL: ParseACL?
+
+    //: Your own properties.
+    var score: Int = 0
+}
+
+//: It's recommended to place custom initializers in an extension
+//: to preserve the convenience initializer.
+extension GameScore {
+
+    init(score: Int) {
+        self.score = score
+    }
+
+    init(objectId: String?) {
+        self.objectId = objectId
+    }
+}
+
 //: Roles can provide additional access/security to your apps.
 
 //: This variable will store the saved role.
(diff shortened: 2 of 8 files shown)


jaysonng:
What's the reason for changing ParseRelation protocol from Codable to Encodable?  (Removing Decodable?)

I'm asking because I now have to implement custom init(from:) for my ParseObjects which has a ParseRelation object nested inside them.

cbaker6:
What’s the usecase for decoding a ParseRelation? You currently can’t fetch this type from the server, why would you fetch it from storage?

jaysonng:
> What’s the usecase for decoding a ParseRelation? You currently can’t fetch this type from the server, why would you fetch it from storage?

In may case, I store a Location object in UserDefaults. (so I have a default Location loaded up in the app without having to wait and find the device current location or have the user select a location.)

The Location object has 2 ParseRelation properties to 2 other ParseObjects -> Tags and Images. 

I can do away saving the images in UserDefaults as they will be huge, but Tags might be something I want stored as well.

cbaker6:
A `ParseRelation` type isn't intended to be decoded. In fact, the only keys that have ever been encoded are: https://github.com/parse-community/Parse-Swift/blob/b92667fb177bbefd59fb79f05e5b1707bc7b501b/Sources/ParseSwift/Types/ParseRelation.swift#L66-L69

This means that even if you decoded a `ParseRelation` in the past, your relation wasn't usable unless you set the `parent` property manually. If you want to decode what you have stored in your `UserDefaults` (and get a working `ParseRelation`), you can create a simple struct with the same keys that are being encoded above:

```swift
struct StoredParseRelation: Decodable {
    var className: String
    var key: String

    // Call this function to get a working ParseRelation on an instance of the struct.
    func createParseRelation<T: ParseObject>(_ parent: T) -> ParseRelation<T> {
        ParseRelation(parent: parent, key: key, className: className)
    }
}
```

jaysonng:
I'll rework my code to get this working. and no, I haven't been able to get to the part of actually checking the ParseRelation objects in my ParseObject before but all these errors popped up when I updated my ParseSwift to 3.1.0 so I had to tackle them now. :)

Thank you once again! I think I understand this more now.

jaysonng:
@cbaker6 ,

**Should we never place ParseRelation properties in structs?**

So whenever a struct that has a ParseRelation, I get the "does not conform to protocol 'Decodable' error.

Screen Shot 2022-01-23 at 5 43 39 PM

It seems like we don't really need to have ParseRelation properties in the struct do we? (Even if the relation is in the Parse database). Was the intention of ParseRelation not be part of the struct object itself? 

From what I understand, I don't access that property at all when I want to query relations anyway right? Otherwise I'd need to be implementing init(from:) on all objects with ParseRelations.

Thank you very much!

cbaker6:
> It seems like we don't really need to have ParseRelation properties in the struct do we? (Even if the relation is in the Parse database). Was the intention of ParseRelation not be part of the struct object itself?

> From what I understand, I don't access that property at all when I want to query relations anyway right? Otherwise I'd need to be implementing init(from:) on all objects with ParseRelations.

An encoded ParseRelation doesn’t have enough info to be useful when decoding. Instead, you should be querying the relatiom like the example in the Playgrounds: https://github.com/parse-community/Parse-Swift/blob/c119033f44b91570997ad24f7b4b5af8e4d47b64/ParseSwift.playground/Pages/12%20-%20Roles%20and%20Relations.xcplaygroundpage/Contents.swift#L281-L343

PR #295, opened 2021-12-03
fix: when transactions enabled throw errors when needed

cbaker6:

### Issue Description

When using transactions, issues can arise when:

1. The number of objects in the transaction are larger than the batch size
2. When child objects haven't been saved

Currently the SDK:

1. Adjusts the batch size to the size of the objects in the transaction. This can be problematic when the number is large
2. Saves unsaved children as individual transactions. If parents can't save after, the original transaction can't be cancelled 

Related issue: #n/a

### Approach

Implement the suggestions in https://github.com/parse-community/Parse-SDK-JS/pull/1090#issuecomment-604225149

### TODOs before merging

- [x] Add tests

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,7 +6,7 @@ coverage:
   status:
     patch:
       default:
-        target: 74
+        target: auto
     changes: false
     project:
       default:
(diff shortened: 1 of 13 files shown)


RahulLanjewar93:
@cbaker6 Hi, just wanted to know what are we doing exactly here to solve both issues.
I'll try to do the same in Parse-JS-SDK

PR #299, opened 2021-12-11
feat: add create(), replace(), and update() to ParseObjects

cbaker6:

### Issue Description

Currently developers have to use `save()` or `saveAll()` to `POST/PUT` objects on a Parse Server. Depending on the configuration, save and saveAll determines what command to send to the server based on the `objectId`, `createdAt`, or a combination of both.

There should be methods to allow the developer to either `POST` or `PUT` where the developer can handle any errors that may come up. This would make it easier for the developer to use a mixed environment of generating objectId’s on the client and server for different objects.

Related issue: #n/a

### Approach

Add `create()` (POST), `replace()` (PUT), and `update()` (PATCH), `createAll()`, `replaceAll()`, and `updateAll()` to all `ParseObject`'s. Follows method discussed in https://github.com/parse-community/Parse-Swift/issues/220#issuecomment-907857645. `update()` and `updateAll()` are marked as internal methods until the server supports PATCH.

Note: Only asynchronous methods using a completion handler, Combine, and async/await are added. No synchronous methods will be added as the developer should be using asynchronous for network calls.

### TODOs before merging

- [x] Refactored ParseObject, ParseUser, and ParseInstallation to reuse more code when using completion callbacks
- [x] Make .POST compatible with current version of .POST on server as well as spec compliant .POST
- [x] Make SDK ready for .PATCH and as http spec compliant as possible while still working with current versions of the Parse Server. All PATCH calls are marked as “internal” and are not usable by developers since PATCH isn't supported on the server yet.  This is a continuation on #146 
- [x] Add tests
- [x] Add entry to changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 [Full Changelog](https://github.com/parse-community/Parse-Swift/compare/2.4.0...2.5.0)
 
 __Improvements__
-- Added create(), update(), createAll(), updateAll() to ParseObjects ([#299](https://github.com/parse-community/Parse-Swift/pull/299)), thanks to [Corey Baker](https://github.com/cbaker6).
+- Added create(), replace(), update(), createAll(), replaceAll(), and updateAll() to ParseObjects. Currently, update() and updateAll() are unavaivalble due to limitations of PATCH on the Parse Server ([#299](https://github.com/parse-community/Parse-Swift/pull/299)), thanks to [Corey Baker](https://github.com/cbaker6).
 - Added convenience methods to convert ParseObject's to Pointer<ParseObject>'s for QueryConstraint's: !=, containedIn, notContainedIn, containedBy, containsAll ([#298](https://github.com/parse-community/Parse-Swift/pull/298)), thanks to [Corey Baker](https://github.com/cbaker6).
 
 ### 2.4.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,15 +2,22 @@
 
 ### main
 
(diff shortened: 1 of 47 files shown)


cbaker6:
@lsmilek1 this should make the issues brought up in https://github.com/parse-community/Parse-Swift/issues/220 about  custom objectId easier to use

lsmilek1:
@cbaker6 , I still need to have `requiringCustomObjectIds: true` in `ParseSwift.initialize` else the server returns `.Not Found` error when I use `.create()` with objectId set for not yet existing object:


> profile from device could not be uploaded, error saving: ParseError code=-1 error=Not Found, profile: Profile ({"ACL":{"*":{"read":true},"VZjyzUPgRk":{"read":true,"write":true}},"ag":0,"bf":258787,"co":1,"d":false,"da":2,"ed":5,"gn":0,"idr":3,"ih":0,"ik":4,"ism":3,"isp":3,"ll":[],"nl":[],"o":1,"objectId":"VZjyzUPgRk","on":false,"ph":[null,null],"pk":{"__type":"Bytes","base64":"XGBGirHaNNfW8LQPG7PH7pa8jqoT2xckPpc+ets8C38="},"pl":0,"po":0,"pt":[null,null],"rl":0,"s":2,"sh":true,"sl":[],"sq":[],"th":0,"tn":1,"tr":1,"tv":0,"tz":"Europe\/Berlin"})

But in that case I would still need to use `ignoringCustomObjectIdConfig` on all `.save()`

Parse server 6.2.0, Parse-Swift 4.14.2

PR #263, opened 2021-10-21
WIP: Strip out unchanged keys on object update

dblythy:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).

### Issue Description

Draft PR for ideas / suggestions for improvements around this approach.

If this isn't suitable for this project, I'll be happy to work on this branch for my own usage. I would welcome feedback so I can make sure my fork is as good as it can be 😊 

Related issue: #242

### Approach

Stores objects in JSON form in memory. When objects are saved, adds unchanged keys to SkipKeys.

### TODOs before merging

- [ ] Add tests
- [ ] Add entry to changelog

PR #315, opened 2022-01-10
feat: add ability to merge updated objects with original

cbaker6:

### Issue Description

The Parse Server doesn't support `PATCH`, but has a custom implementation of `PUT` which allows the client to only send a subset of keys that need updating. This custom implementation "acts" as a `PATCH`. Seem more in #242

In order for the Swift SDK to take advantage of the Parse Servers custom implementation of `PUT`; to act as a `PATCH`, developers need to make their `ParseObject`'s conform to `ParseObjectMutable` and use `.mutable` on their `ParseObject` before changing any keys. 

***Issue:*** When a developer uses `.mutable` after conforming to `ParseObjectMutable`, the returned result from a save or update only returns the `updated` object, not the updated+original object, meaning the returned object on the client doesn't have the same keys as the server. In order to retrieve the full updated object, the developer will have to merge the original and updated themselves or in the case of `ParseUser.current` and `ParseInstallation.current`, they have to fetch the updated object from the cloud.  A related PR is #263.

If a developer isn't using `.mutable`, it means they are always sending all keys to the server and are essentially doing a true `PUT` (without upsert, as the Server doesn't support upsert), so this problem isn't relevant in those situations. Similarly, the problem also isn't relevant when creating objects using `POST`.

Related issue: #242 

### Approach

Add protocol implementations and helper methods that allow developers to specify all of their keys and allow for the original and updated objects to be merged by using `updated.merge(original)`.  ~~This will need to be called by objects that aren't stored in the keychain by the Swift SDK. Keychain objects (those that have `.current`) have their changes merged by the SDK automatically.~~ This makes the client object match the server object after an update automatically assuming the developer used the renamed `mutable->mergeable` property on a `ParseObject` before making the modification and that they implemented the new `merge()` method.

The following protocol properties need to be added to ALL `ParseObject`'s by ALL developers, resulting in a breaking change:
```swift
 /**
     A JSON encoded version of this `ParseObject` before `mergeable` was called and
     properties were changed.
     - warning: This property is not intended to be set or modified by the developer.
*/
    var originalData: Data? { get set }
```

The following method developers CAN(not required) add to their `ParseObjects` if they want to take advantage of merging:

```swift
/**
      Merges two `ParseObject`'s with the resulting object consisting of all modified
     and unchanged properties.

          //: Create your own value typed `ParseObject`.
              //: These are required by ParseObject
              var objectId: String?
              var createdAt: Date?
              var updatedAt: Date?
              var ACL: ParseACL?
              var originalData: Data? // New property that the compiler will enforce to be added

              //: Your own properties.
              var points: Int?

              //: Implement your own version of merge
              func merge(_ object: Self) throws -> Self {
                  var updated = try mergeParse(object)
                  if updated.shouldRestoreKey(\.points,
                                                   original: object) {
                      updated.points = object.points
                  }
                  return updated
              }
          }

      - parameter object: The original object.
      - returns: The merged object.
      - throws: An error of type `ParseError`.
      - important: It is recommend you provide an implementation of this method
      for all of your `ParseObject`'s as the developer has access to all keys of a
      `ParseObject`. You should always call `mergeParse`
      in the beginning of your implementation to handle all default Parse keys. In addition,
      use `shouldRestoreKey` to compare key modifications between objects.
 */
func merge(_ object: Self) throws -> Self
```

The following protocol helper methods are now available for developers to use to assist with their implementations of `merge()`:

```swift
 /**
      Merges two `ParseObject`'s resulting in modified and unchanged Parse keys.
      - parameter object: The original installation.
      - returns: The updated installation.
      - throws: An error of type `ParseError`.
      - warning: You should only call this method and shouldn't implement it directly
      as it's already implemented for developers to use.
*/
func mergeParse(_ object: Self) throws -> Self

/**
      Determines if a `KeyPath` of the current `ParseObject` should be restored
      by comparing it to another `ParseObject`.
      - parameter original: The original `ParseObject`.
      - returns: Returns a `true` if the keyPath should be restored  or `false` otherwise.
*/
func shouldRestoreKey<W>(_ key: KeyPath<Self, W?>, original: Self) -> Bool where W: Equatable
```

An example of how this works can be found in the Playgrounds (notice that merge doesn't need to be called by the developer because merges happen automatically: https://github.com/parse-community/Parse-Swift/blob/d180c633525d54de1372a5d07171ddef104deae5/ParseSwift.playground/Pages/6%20-%20Installation.xcplaygroundpage/Contents.swift#L16-L84

An example with `ParseObject` can be seen in Playgrounds:

https://github.com/parse-community/Parse-Swift/blob/e12b88f13fe02c0a46bc97fc235d8220f8c3841a/ParseSwift.playground/Pages/1%20-%20Your%20first%20Object.xcplaygroundpage/Contents.swift#L32-L66

~I might be able to leverage the current implementation to remove the need for the developer to implement `merge`, but I need to think about it a little more. Basically the developer will need to give the SDK a list of all KeyPaths (for each `ParseObject` to remove a `merge` implementation requirement.~ Update: I wasn't able to come up with a way to make this feasible with key paths.

### Notes
~~The current changes are non-breaking and will be released as a minor update. Later in the year, a major release will occur requiring all keys either be optional for ParseObjects or to have a default value.~~ The current changes are breaking (weak break) and will need to be released as a major update. The compiler will require all ParseObject's to add `var originalData: Data?`. 

In addition, all `ParseObject`'s have `mergeable` by default. Essentially reverting back #243 to which prevents developers from setting up their `ParseObject`'s incorrectly by not using optionals for all properties and allowing the SDK to be able to initialize new versions of their `ParseObject`'s when necessary. Developers should NOT attempt to enforce properties be defined by defining their property as a non-optional type. Instead, a developer should do this by creating additional initializers (preferably in an extension) and carry out any necessary checks. Enforcing properties to be defined can also be done on the server side via schema setup, Parse Dashboard, or Cloud Code.

### TODOs before merging

- [x] Add tests
- [x] Add entry to changelog


cbaker6:
@dblythy @vdkdamian Since you all were part of the discussion about only saving keys of updated objects, can you take a look at this PR to provide your thoughts? Any opinions with naming and how it works, etc. will be helpful...

vdkdamian:
> @dblythy @vdkdamian @Vortec4800 Since you all were part of the discussion about only saving keys of updated objects, can you take a look at this PR to provide your thoughts? Any opinions with naming, how it works, design, etc. will be helpful...
> 
> 
> 
> I will add more details to description later today/tomorrow.

I'll take a look at this thursday.

cbaker6:
I'm thinking of changing the name of one of the added methods, `isRestoreOriginalKey -> isShouldRestoreKey`

mtrezza:
> notice that merge doesn't need to be called by the developer because ParseInstallation.current merges happen automatically

Is it possible to add that behavior as default for all Parse Objects?

> Later in the year, a major release will occur... 
That's a good reminder that we make the Swift SDK conform to the broader Parse release cycle and deprecation policy sometime this year.

cbaker6:
> Is it possible to add that behavior as default for all Parse Objects?

It’s not possible in the current context because the SDK doesn’t have the original object. The developer would have pass the original ParseObject to the save method in order for it to be merged automatically.

vdkdamian:
This definitely is an improvement to fix the current issue, but I do agree that it's making things a little complicated for some developers that may not be following the conversations we had. 

If I have some time within a few weeks I'll try to help look for a better solution, but for now this definitely seems like the best way to do things.

> I'm thinking of changing the name of one of the added methods, `isRestoreOriginalKey -> isShouldRestoreKey`

It still seems a little weird to me, what about just using `shouldRestoreKey`? Is there a reason why u decided to put "is" before it?

cbaker6:
> but I do agree that it's making things a little complicated for some developers that may not be following the conversations we had.

This happens with any repo. Each developer is not expected to keep up with each PR a repo opens/merges. For devs already using the SDK in it's current form and not using `ParseMutable` probably won't care about this PR and won't need to follow along. They will just need to add `var originalObject: Data?` to their `ParseObject`'s if they choose to upgrade and use the SDK as they have been doing.

The PR is only important to those who would like to use `PATCH` (or in the Parse Servers current context, a quirky version of `PUT` that acts as a `PATCH`). For these devs, they will want to read the README, release notes, and Playgrounds to get this functionality to add `merge` to their specific objects. If these devs don't do this, the SDK will act as I noted here: https://github.com/parse-community/Parse-Swift/pull/315#issue-1098220571

A developer familiar with REST will expect the Swift SDK to do what it's been doing in all previous versions of the Swift SDK.

mtrezza:
> but I do agree that it's making things a little complicated for some developers that may not be following the conversations we had.

I agree. Another major user group are developers who are just getting started with Parse Swift SDK,
- who are either entirely new to Parse, maybe coming from a 3rd party like [Firebase](https://firebase.google.com/docs/firestore/manage-data/add-data#update-data)
- or who are already using other Parse client SDKs and are expecting consistent behavior among different Parse SDKs.

cbaker6:
> Maybe it's worth to compare the approach to how similar frameworks solve this in their Swift SDKs and aiming for a consistent simplicity for updating objects among Parse SDKs.

Can you enlighten us on what/who are these "similar frameworks" who have "Swift SDKs" and explain why you believe they are similar?

dblythy:
@cbaker6 I regret to inform that I'm not currently available to provide any constructive feedback, however I am excited and grateful for your efforts to ensure consistency/clarity across various SDKs

mtrezza:
> Can you enlighten us on what/who are these "similar frameworks" who have "Swift SDKs" and explain why you believe they are similar?

Firebase or any other framework that can be considered an alternative to Parse.

cbaker6:
> Firebase or any other framework that can be considered an alternative to Parse.

Firebase is actually similar to the [Parse Objective-C SDK](https://github.com/parse-community/Parse-SDK-iOS-OSX). If you would like to "follow" Firebase's route, you should create PR's and provide suggestions to the Objective C SDK that add Swift files and extensions to Objective C files.

The Parse Swift SDK is not similar to Firebase by any means...

mtrezza:
I am talking about SDK usage complexity and consistency with other Parse SDKs in the sense of language agnostic, developer friendly design choices.

cbaker6:
>  If something is perceived as complex in the Swift SDK by other users, we can look into how we could address these concerns in the Swift SDK, not in the ObjC SDK.

Have you used the Swift SDK before? If so, can you explain what is complex in the usage of this SDK, this PR, and why you feel it's complex?

cbaker6:
> I regret to inform that I'm not currently available to provide any constructive feedback, however I am excited and grateful for your efforts to ensure consistency/clarity across various SDKs

@dblythy np, thanks for PR 263, the design an discussion on that thread influenced how I went about the `originalData` property, so I definitely appreciate your effort

mtrezza:
> Have you used the Swift SDK before? If so, can you explain what is complex in the usage of this SDK, this PR, and why you feel it's complex?

Sure, see https://github.com/parse-community/Parse-Swift/pull/315#issuecomment-1012503764 and https://github.com/parse-community/Parse-Swift/issues/242.

cbaker6:
> Sure, see #315 (comment) and #242.

You never answered the questions: 

1. Have you used the Swift SDK before?
2. If so, can you explain what is complex in the usage of this SDK, this PR

It seems, you "feel it's complex" because of someone else's comments. "implementing" a method to do something more specific to a `struct`'s need is no more "complex" than overriding a `class` method. A developer familiar with classes and reference types should have no problem with this, particularly when they are looking for very "specific" functionality like only sending updated keys.

mtrezza:
I answered your question about complexity in this SDK. The issue of complexity has been discussed thoroughly in the other thread already where I had to jump in as the discussion was going off the rails with personal attacks. Not sure why you want to continue here, but the pattern seems to repeat and I'm not available for that.

cbaker6:
@pmmlo since it looks like you might be available again and provided a lot feedback in the early stages of this SDK, do you mind looking at the description of this PR? Mainly if it makes since to you from a Swift and POP standpoint and do the Swift Playgrounds seem intuitive? Also, your thoughts on the new variable and method names.

CHANGELOG.md
```diff
@@ -18,7 +31,7 @@ __New features__
...
 - Adds isNull and isNotNull QueryConstraint along with the ability set/forceSet null using ParseOperation ([#308](https://github.com/parse-community/Parse-Swift/pull/308)), thanks to [Corey Baker](https://github.com/cbaker6).
 - Adds auth support for GitHub, Google, and LinkedIn ([#307](https://github.com/parse-community/Parse-Swift/pull/307)), thanks to [Corey Baker](https://github.com/cbaker6).
-- (Breaking Change) Adds options to matchesText QueryConstraint along with the ability to see matching score. The compiler should recommend the new score property to all ParseObjects ([#306](https://github.com/parse-community/Parse-Swift/pull/306)), thanks to [Corey Baker](https://github.com/cbaker6).
+- (Breaking Change) Adds options to matchesText QueryConstraint along with the ability to see matching score. The compiler will recommend the new score property be added to all ParseObjects ([#306](https://github.com/parse-community/Parse-Swift/pull/306)), thanks to [Corey Baker](https://github.com/cbaker6).
```
pmmlo, line 34: - (Breaking Change) Adds options to matchesText QueryConstraint along with the ability to see matching score. The compiler will recommend the new score and originalData properties be added to all ParseObjects ([#306](https://github.com/parse-community/Parse-Swift/pull/306)), thanks to [Corey Baker](https://github.com/cbaker6).

Sources/ParseSwift/Objects/ParseObject.swift
```diff
@@ -84,6 +175,28 @@ public extension ParseObject {
     func toPointer() throws -> Pointer<Self> {
         return try Pointer(self)
     }
+
+    func shouldRestoreKey<W>(_ key: KeyPath<Self, W?>,
+                             original: Self) -> Bool where W: Equatable {
+        self[keyPath: key] == nil && original[keyPath: key] != self[keyPath: key]
```
pmmlo, line 181: @cbaker6 You spoil everyone with these methods.  I vaguely remember someone wanted to set values as `nil`.  This is really elegant.

pmmlo:
> @pmmlo since it looks like you might be available again and provided a lot feedback in the early stages of this SDK, do you mind looking at the description of this PR? Mainly if it makes since to you from a Swift and [POP](https://www.pluralsight.com/guides/protocol-oriented-programming-in-swift) standpoint and do the Swift Playgrounds seem intuitive? Also, your thoughts on the new variable and method names.

I think it looks great.

tl;dr

Appreciate you pinging me on this.  This looks incredibly powerful, because you can delve into both the original and updated objects with the key paths and manually determine how to reconcile each key.

To me, it looks very intuitive for a pretty nuanced challenge.  I think the naming is excellent.  I skipped over the tests.  Your playgrounds always look clear to me.

I bias toward classical and old school, so I really like how you implemented this, though it is a breaking change.  But I suppose if you insisted on going as POP as possible with backward compatibility you could do something like this:
```
/// This implementation may not be exactly right because I just wrote this off the cuff in plaintext.
///  Hopefully you get the idea.
struct Foo: ParseObject, ParseRestorable {
    var objectId: String?
    var createdAt: Date?
    var updatedAt: Date?
    var ACL: ParseACL?
    var score: Double?

    var bar: Int?
    
    var originalData: Data? /// <-- Enforced by ParseRestorable
     func merge(_ object: Self) throws -> Self {
        var updated = try mergeParse(object)
        if updated.shouldRestoreKey(\Foo.customKey, original: object) {
            if object.customKey == nil {
                updated.customKey = 0
            } else { updated.customKey = object.customKey }
        }
    }   
}

public protocol ParseRestorable: Objectable, Fetchable, Savable, Deletable, Identifiable, Hashable {
    var originalData: Data? { get set }
    var mutable: Self { get }

   func shouldRestoreKey<W>(_ key: KeyPath<Self, W?>, original: Self) -> Bool where W: Equatable

    func mergeParse(_ object: Self) throws -> Self

    func merge(_ object: Self) throws -> Self

    init()
}

public extension ParseRestorable {
/// Base implementations.
}
```

I suppose if we wanted to be doctrinal about POP, you could do the above and get the extra granularity without a breaking change, but I think your implementation is far superior for a number of reasons.

vdkdamian:
> I suppose if we wanted to be doctrinal about POP, you could do the above and get the extra granularity without a breaking change, but I think your implementation is far superior for a number of reasons.

Since I said before that it's making things a little complicated for some developers, I really like your idea. Seems good to add a protocol that will add the function and variable to your own struct.

To me, this all looks really good!

pmmlo:
> Since I said before that it's making things a little complicated for some developers, I really like your idea. Seems good to add a protocol that will add the function and variable to your own struct.
> 
> To me, this all looks really good!

IMHO, I think we should go with the current state (e.g. allow devs to overload the merge methods, up the version to a major release (4.0), and require devs wanting to use Parse-Swift >4.0 to add `var originalData: Data?`).


tl;dr

I think it's great there is thought for new adoption and controlling complexity.

So, is the complexity around keyPath?  My guess would be that concept and reconciling merges would be most challenging for new devs.  But, like @cbaker6 points out, the only new requirement is `var originalData: Data?`, which new devs are going to rely on Xcode to autofill anyways (along with objectId, createdAt, updatedAt, ACL, and score).

I bias toward classical OOP over POP.  In this instance, by creating an additional protocol, we are essentially penalizing good coding practice by requiring additional struct type to every model.

I have always believed creating protocols for everything only makes the code base more verbose, not simpler (because you have to remember all of the various type declarations and their individual methods).  I think if anything `score` should have been a separate protocol or a subclass of `ParseObject` because people may not want to use `matchesText` which is not a very efficient operation and not every model is going to have a `String`.

I also think by requiring `originalData` some new devs will be prompted to learn what it's used for and learn to optimize data transfer, while a separate operation would require "documentation diving" and awareness the protocol's existence.

If the community consensus is still to have an additional protocol, I think we should at least spend some time on the naming.  `ParseRestorable` seems to be a bad option.  I think `ParseMergeable` or `ParseReconcilable` are a little better, but I don't think they capture the idea of optimizing and update.

Another option I didn't see anybody bring up is subclassing from 'ParseObject' to make something like 'AdvancedParseObject', but the naming becomes really tricky because it could become a catch-all over time.

Sources/ParseSwift/Objects/ParseObject.swift
cbaker6: Haha, I try to make it as easy as possible to use. Plus, as you know, leveraging `KeyPath`'s typically makes things a lot easier for the end-user/developer.

CHANGELOG.md
cbaker6: Here I was fixing some grammar from an older PR. I put the `originalData` info in the `4.0.0` part of the changelog above

cbaker6:
@vdkdamian @pmmlo thanks for looking this over and providing great feedback!

pmmlo, below are responses to your feedback and suggestions:

> I bias toward classical and old school, so I really like how you implemented this, though it is a breaking change. But I suppose if you insisted on going as POP as possible with backward compatibility you could do something like this:

> I suppose if we wanted to be doctrinal about POP, you could do the above and get the extra granularity without a breaking change, but I think your implementation is far superior for a number of reasons.

By classical OOP, I'm assuming you mean grouping multiple protocols into `ParseObject` so the developer essentially inherits all of the abilities of all of the Protocols at once as oppose to having to add each one individually? If I understood you correctly, I agree with you. This allows every developer who has a `ParseObject` to take advantage of the updates without having to do much work because the compiler suggests whatever update they need.

Using `ParseRestorable` is definitely a possibility without causing a break, but then the developer has to know about this protocol and decide when to add it to a `ParseObject` themselves. In addition, if we plan on reconciling this protocol into `ParseObject` in a future update, it will eventually cause another breaking change (or warning).

Vortec4800:
Thanks for pinging me on this, sorry I couldn't spend some time on it sooner.

I think this is a really good solution for a really complex problem. While I do worry in general about adding complexity, this does seem like it's mostly optional - and adding an additional key isn't a huge lift. I also think the `originalData` property is well-named, however I have the insight from following along our conversations so we may want some portion of documentation that talks about the issue, why this is here, why you would want to use it, etc. I haven't had a chance to go through the Playground yet so perhaps it's in there already.

If there is one thing that worries me it's this:

> Developers should NOT attempt to enforce properties be defined by defining their property as a non-optional type. Instead, a developer should do this by creating additional initializers (preferably in an extension) and carry out any necessary checks. Enforcing properties to be defined can also be done on the server side via schema setup, Parse Dashboard, or Cloud Code.

I do really like having the Parse object be the source of truth when it comes to vars that are required vs vars that aren't, and having that be enforced by the compiler. I understand the need, these "thin" objects will need to have empty values, but it's an unfortunate side effect. I don't have a solution here, without making a separate thin version with all optional properties, but that would be quite a bit of extra work to maintain. I guess in the end it's fine, and we'll just have to enforce it in other ways as you mentioned.

> If the community consensus is still to have an additional protocol, I think we should at least spend some time on the naming. `ParseRestorable` seems to be a bad option. I think `ParseMergeable` or `ParseReconcilable` are a little better, but I don't think they capture the idea of optimizing and update.

I agree on this as well. While I understand why `ParseRestorable ` makes sense, I also don't think it immediately conveys the reason to use it. Perhaps something like `ParseOptimizedSendable` or something in that world that encapsulates the functionality you would gain by implementing it. If I was a developer only skimming half the documentation (I'm sure that never happens...) and I saw something called `ParseRestorable` that wasn't required, I would assume it had something to do with state restoration or something and skip it entirely.

I really like that this is optional. The apps my team makes that use Parse are often quite small, and don't use that much traffic. I think for most of the apps, we'd actually skip this protocol and keep things simple given we don't need the speed improvement this would provide on data transmission. That said we do also have a couple very large projects that run on load balancers and clusters, and this would be a big improvement for those and worth the extra work and thought to implement.

cbaker6:
@Vortec4800 thanks for your feedback!

> Developers should NOT attempt to enforce properties be defined by defining their property as a non-optional type. Instead, a developer should do this by creating additional initializers (preferably in an extension) and carry out any necessary checks. Enforcing properties to be defined can also be done on the server side via schema setup, Parse Dashboard, or Cloud Code.

> I do really like having the Parse object be the source of truth when it comes to vars that are required vs vars that aren't, and having that be enforced by the compiler. I understand the need, these "thin" objects will need to have empty values, but it's an unfortunate side effect. I don't have a solution here, without making a separate thin version with all optional properties, but that would be quite a bit of extra work to maintain. I guess in the end it's fine, and we'll just have to enforce it in other ways as you mentioned.

There are 3 important reasons for this that many might not realize when they chose to go the non-optional route:

1. This isn't inline with the standard JSON Swift encoder/decoder and how it uses optionals. You and I discussed this in a previous thread
2. How Parse pointers work and how Parse uses them. Parse Pointers ONLY have the following properties: `className`, `objectId`, and `__type`. Since the Swift SDK is heavily dependent on generics, a developer making a key required that isn't in the aforementioned list makes it impossible to create `ParsePointer`'s from their `ParseObject`'s. Though someone may argue they don't need Pointer's in the current version of their app, they will most likely them when their app grows 
3. From a POP standpoint, if the `ParseObject` protocol doesn't define `init()`, developers can make non-optional values and it's impossible to initialize new versions of their ParseObject's internally in the SDK; preventing `.mutable`, `merge`, or anything else that needs a new copy inside the app. For value types with all optional properties, the compiler provides the `init()` automatically; assuming all other inits are defined in an extension.

Originally, It didn't cross my mind that developers would create `ParseObject`'s with non-optional properties since every property given to them from `ParseObject` is optional. I mentioned on a previous thread that if I would have realized developers wouldn't follow the design of a `ParseObject` given to them I would have enforced an `init()` during the first release.

cbaker6:
@pmmlo @vdkdamian @Vortec4800 For the `ParseObject` property, I'm thinking this is a better name:

- `.mutable -> .mergable`

Thoughts?

vdkdamian:
> For the `ParseObject` property, I'm thinking this is a better name:
> 
> * `.mutable -> .mergable`

Yes, I definitely agree

pmmlo:
Could we do `mergeable`?

> @pmmlo @vdkdamian @Vortec4800 For the `ParseObject` property, I'm thinking this is a better name:
> 
>     * `.mutable -> .mergable`
> 
> 
> Calling it will look like:
> 
> ```swift
> /*: Update your `ParseInstallation` `customKey` value. 
>      Performs work on background queue and returns to designated on 
>      designated callbackQueue. If no callbackQueue is specified it 
>      returns to main queue. 
>   */ 
>  var installationToUpdate = Installation.current?.mergable // Name change here.  
>  installationToUpdate?.customKey = "myCustomInstallationKey2" 
>  installationToUpdate?.save { results in 
>   
>      switch results { 
>      case .success(let updatedInstallation): 
>          print("Successfully save myCustomInstallationKey to ParseServer: \(updatedInstallation)") 
>      case .failure(let error): 
>          print("Failed to update installation: \(error)") 
>      } 
>  } 
> ```
> 
> Thoughts?

vdkdamian:
> Could we do `mergeable`?

Yes, this is the correct one, didn't notice it.

cbaker6:
@pmmlo good catch!

PR #319, opened 2022-01-16
fix: move score property into ParseQueryScorable

cbaker6:

### Issue Description

Swift SDK 3.0.0 made all `ParseObjects`'s have to have the property `var score: Double?` though this property is only needed when using the `matchesText` `QueryConstraint`. 

Related issue: #n/a

### Approach

Move the `var score: Double?` to a protocol named `ParseQueryScorable`. When developers want to sort by score using a `matchesText` `QueryConstraint`, they just conform their `ParseObject`'s to  `ParseQueryScorable`. Looks like below:

Developers who are not using the score property, but have upgraded can do either of the following:

1. Nothing, if you added the score property already or didn't add it, this fix won't break your code
2. If you don't plan to use the score property, you can remove it from your `ParseObject`'s as it won't break your code

### TODOs before merging

- [x] Modify tests
- [x] Add entry to changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,15 @@
 
 ### main
 
-[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/3.1.1...main)
+[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/3.1.2...main)
 * _Contributing to this repo? Add info about your change here to be included in the next release_
+
+### 3.1.2
+[Full Changelog](https://github.com/parse-community/Parse-Swift/compare/3.1.1...3.1.2)
+
+__Fixes__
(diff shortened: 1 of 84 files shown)


cbaker6:
@pmmlo can you let me know what you think about this?

pmmlo:
You are a rockstar!  Sorry to make more work.  I honestly liked it the way you had it. 

But, if we’re embracing protocols, do you think it would make sense to move 
`public func sortByTextScore() -> Query<T> {}` from Query.swift to ParseQueryScorable.swift to enforce using the `var score` together with `sortByTextScore()`?

I’m on mobile but nothing I could see other than that.

cbaker6:
> But, if we’re embracing protocols, do you think it would make sense to move
public func sortByTextScore() -> Query<T> {} from Query.swift to ParseQueryScorable.swift to enforce using the var score together with sortByTextScore()?

Not sure how to make this happen since sortByTextScore modifies the query and not the ParseObject. Do you have an idea on how to cary this out?

I think it’s okay if sortByTextScore remains in query as essentially some can sort and never look at score.

cbaker6:
Thinking about this more, I may be able to add:

```swift
public func sortByTextScore() -> Query<T> where T: ParseObject & ParseQueryScorable {
```

but this can possibly cause a breaking change to anyone who began using `sortByTextScore()` in 3.0.0. So, it's probably best to that method untouched for now.

Issue #313, opened 2022-01-08
Response parsing error when ParseUser is successfully signed up


Climbatize:
### New Issue Checklist

- [X] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [X] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [X] I have searched through [existing issues](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).
- [X] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Swift SDK](https://github.com/parse-community/Parse-Swift/releases). 

### Issue Description

When sining up a User Parse sever response may not contain a sessionToken (Parse 4.2)

### Steps to reproduce

- Create a ParseUser subclass

```swift
import Foundation
import ParseSwift
​
struct User: ParseUser {
    var emailVerified: Bool?
    var password: String?
    var authData: [String: [String: String]?]?
    var createdAt: Date?
    var updatedAt: Date?
    var ACL: ParseACL?
​
    public var objectId,
			   email,
               qrCode:         String?
​
​
}
```

- Uses one of the ParseUser methods to sign up an instance ()

```swift
var user = User();
                   user.password = password;
                   user.username = username;
                   user.email = email;
                   user.qrCode = "";
            
user.signup() { result in
    switch result {
    case .failure(let error):
        //: Publish error.
        self.error = error
        print(error.message)
    case .success(user):
        print("Successfully queried data")
    case .success(_):
        print("Successfully queried data")
    }
}
```
or
```swift
var user = User();
                   user.password = password;
                   user.username = username;
                   user.email = email;
                   user.qrCode = "";
            
let publisher = user
    .signupPublisher()
        .sink(
                receiveCompletion: { result in
                    switch result {
                    case .failure(let error):
                        //: Publish error.
                        self.error = error
                        print(error.message)
                    case .finished:
                        print("Successfully queried data")
                    }
                }, receiveValue: { user in
                    self.user = user
                }
        )
publisher.store(in: &subscriptions)
```

### Actual Outcome
Subscription results as an unknown error but the ParseUser is stored in database

### Expected Outcome
Returns a success or an explaination on why the outcome is an error

### Environment

Client
- Parse Swift SDK version: `3.0.0`
- Xcode version: `13.2.1`
[...]

PR #331, opened 2022-01-24
fix: sessionToken nil if preventLoginWithUnverifiedEmail=1 on server

Climbatize:

- [X] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [X] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).

### Issue Description
When Parse server is configured with `preventLoginWithUnverifiedEmail` session token after signup is nil and causes a parsing crash, caught and returned as unknown error.

Related issue: #313

### Approach
What I do there is setting the sessionToken variable in signup response to optional to prevent parsing crash. Then, if the token is nil, I throw a 209 so developers on SDK can identify easily the ParseError and provide the good message to their users.

### TODOs before merging

- [ ] Add entry to changelog

### DIFF
--- a/Sources/ParseSwift/Objects/ParseUser.swift
+++ b/Sources/ParseSwift/Objects/ParseUser.swift
@@ -245,9 +245,11 @@ extension ParseUser {
         return API.Command<SignupLoginBody, Self>(method: .POST,
                                                   path: .login,
                                                   body: body) { (data) -> Self in
-            guard let sessionToken = try ParseCoding.jsonDecoder().decode(LoginSignupResponse.self, from: data).sessionToken
+            guard
+                let sessionToken = try ParseCoding.jsonDecoder()
+                    .decode(LoginSignupResponse.self, from: data).sessionToken
             else {
-                throw ParseError(code: .invalidSessionToken, message:"no session token provided by the server")
+                throw ParseError(code: .invalidSessionToken, message: "no session token provided by the server")
             }
             let user = try ParseCoding.jsonDecoder().decode(Self.self, from: data)
 
@@ -270,7 +272,10 @@ extension ParseUser {
      - note: The default cache policy for this method is `.reloadIgnoringLocalCacheData`. If a developer
      desires a different policy, it should be inserted in `options`.
     */
-    public func become(sessionToken: String, options: API.Options = []) throws -> Self {
+    public func become(sessionToken: String?, options: API.Options = []) throws -> Self {
+        guard let sessionToken = sessionToken else {
+            throw ParseError(code: .invalidSessionToken, message: "no session token provided")
+        }
         var newUser = self
         newUser.objectId = "me"
         var options = options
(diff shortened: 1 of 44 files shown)


vdkdamian:
Can I first try this with my project before merging? Just to make sure this isn't breaking anything.
I'll try to do this within the next few days, after @cbaker6 approves these changes.

vdkdamian:
Can you also put this on a different branch?

cbaker6:
Can you provide appropriate testcases along with fixing the tests that are failing?

Sources/ParseSwift/Objects/ParseUser.swift
```diff
@@ -245,7 +245,10 @@ extension ParseUser {
         return API.Command<SignupLoginBody, Self>(method: .POST,
                                                   path: .login,
                                                   body: body) { (data) -> Self in
-            let sessionToken = try ParseCoding.jsonDecoder().decode(LoginSignupResponse.self, from: data).sessionToken
+            guard let sessionToken = try ParseCoding.jsonDecoder().decode(LoginSignupResponse.self, from: data).sessionToken
```
cbaker6, line 248: This should be `try?` or else it will still throw the original decoding error

Sources/ParseSwift/Objects/ParseUser.swift
```diff
@@ -699,8 +702,11 @@ extension ParseUser {
                     path: .users,
                     body: body) { (data) -> Self in
 
-            let sessionToken = try ParseCoding.jsonDecoder()
-                .decode(LoginSignupResponse.self, from: data).sessionToken
+            guard let sessionToken = try ParseCoding.jsonDecoder()
```
cbaker6, line 705: Same here. When you add the necessary test cases, it will show this

cbaker6 (changes requested):

Sources/ParseSwift/Objects/ParseUser.swift
Climbatize: In fact I kept the original behavior, it has to fail if the response is malformed

Sources/ParseSwift/Objects/ParseUser.swift
Climbatize: it will not throw the original decoding error become I set `sessionToken` optional, it will only fail on other parsing errors

cbaker6: I believe the server should throw the error and sessionToken should be left in it's original form. I'll take a look soon...

Climbatize: I totally agree on that fact @cbaker6 , I did not even think we could have a fix on server side 👍

Climbatize:
> Can you also put this on a different branch?

What is your branching model?

Climbatize:
> Can you provide appropriate testcases along with fixing the tests that are failing?
> 
> You also need to lint your branch https://github.com/parse-community/Parse-Swift/blob/main/CONTRIBUTING.md#setting-up-your-local-machine
> 
> What needs to be investigated is if the server or client is responsible for throwing the error. It seems like something like this should come from the server as the client isn’t aware of the server config.

I provided test cases and tests are running well locally, do you know what happening on Github Action?

image

cbaker6:
I'm not able to replicate your issue based on what's been provided so far. I recommend following the directions here: https://github.com/parse-community/Parse-Swift/issues/313#issuecomment-1020310987

> I provided test cases and tests are running well locally, do you know what's happening on Github Action?

Not sure, but if actions is showing these type of errors, there's most likely something wrong with the PR as oppose to actions.

Climbatize:
solved by #332
Seems like it's not today you'll get another contributor from the community :). Thanks @cbaker6 and @vdkdamian for your help

cbaker6:
@Climbatize you found out there's an issue, so you are definitely a contributor!

I'm sure there's something you will add soon to the SDK!

Thanks for reporting!!!

PR #374, opened 2022-06-21
Update .spi.yml

finestructure:
Remove additional documentation targets which are platform flavours.

@cbaker6 Corey, I believe the intention here was to create four different doc sets here, one per platform. Unfortunately, that's not how our DocC generation (or DocC generation in general?) works.

The docset for all four targets is called `parseswift` and they end up simply overwriting each other in sequence. So currently you're hosting the last one, for watchOS. It's an outstanding item on my todo list to bring up with the DocC working group how multi-platform docsets are supposed to be generated and hosted side-by-side. I don't think there's anything we can do at the moment to address this :(

I'm not sure to what extent you have control over creating entirely separate doc targets which would allow you to ship separate sets per platform if they're materially different. If that's the case, would you like to open a discussion over in https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/discussions to see if this can be worked around somehow? I can go into more detail how the process works from our end which might allow experimentation on your end.

If there is one canonical or best suited set of docs to generate, it's best to only specify that for now!

- [ ] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [ ] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).

### Issue Description

Related issue: #`FILL_THIS_OUT`

### TODOs before merging

- [ ] Add tests
- [ ] Add entry to changelog

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -5,7 +5,7 @@ builder:
     scheme: "ParseSwift (iOS)"
   - platform: macos-xcodebuild
     scheme: "ParseSwift (macOS)"
-    documentation_targets: ["ParseSwift (iOS)", "ParseSwift (macOS)", "ParseSwift (tvOS)", "ParseSwift (watchOS)"]
+    documentation_targets: ["ParseSwift (macOS)"]
   - platform: macos-xcodebuild-arm
     scheme: "ParseSwift (macOS)"
   - platform: tvos


cbaker6:
>  I believe the intention here was to create four different doc sets here, one per platform. Unfortunately, that's not how our DocC generation (or DocC generation in general?) works.

Hi Sven, thanks for the fix. I was testing this based on the [SPI blog](https://blog.swiftpackageindex.com/posts/auto-generating-auto-hosting-and-auto-updating-docc-documentation/) with the intent to see what happens. 

> I'm not sure to what extent you have control over creating entirely separate doc targets which would allow you to ship separate sets per platform if they're materially different.

The docs for Pare-Swift are essentially the same, so we can go with one target.

cbaker6 (approved):

Issue #382, opened 2022-07-13
Enable queries to use caching policies


aliasad106:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).

### Current Limitation
The query does not allow me to add cachePolicy cacheThenNetwork as it was implemented there in the Parse iOS/macOS SDK. 

### Feature / Enhancement Description
When given the .cacheThenNetwork. It will first check from cache and respond and then it will load from network.

### Example Use Case
Example use case is. When I load my feeds it first shows from cache and simultaneously fetches from network and respond and it also updates the cache with new data. 

### Alternatives / Workarounds
No 

### 3rd Party References
[Parse](https://docs.parseplatform.org/ios/guide/#caching-queries)

cbaker6:
Please complete the [issue template](https://github.com/parse-community/Parse-Swift/issues/382#issue-1303958596) in it's entirety by adding information under each header.

Caching was added in #196. You can change the default cache policy the SDK uses for your app when you initialize the SDK using `requestCachePolicy`, `cacheMemoryCapacity`, and `cacheDiskCapacity` as stated in the [documentation](https://parseplatform.org/Parse-Swift/release/documentation/parseswift/parseswift/initialize(applicationid:clientkey:masterkey:serverurl:livequeryserverurl:allowingcustomobjectids:usingtransactions:usingequalqueryconstraint:keyvaluestore:requestcachepolicy:cachememorycapacity:cachediskcapacity:migratingfromobjcsdk:deleti-97083)). Details can be found by looking at Apple's [documentation](https://developer.apple.com/documentation/foundation/url_loading_system/accessing_cached_data).

If you want to choose a different caching policy for an individual query, simply pass the caching policy option you want to your query:

```swift
let options: API.Options = [.cachePolicy(.reloadIgnoringLocalCacheData))]
let query = GameScore.query("points" > 50)
let results = try await query.find(options: options)
...
```

cbaker6:
[...]

PR #386, opened 2022-08-03
feat: enable query caching by using GET instead of POST

cbaker6:

### Issue Description

All queries currently use `POST` requests, preventing the ability to leverage the cache policies available in the SDK.

Related issue: #382 

### Approach

Enable all queries to use `GET` requests by default so queries are cached. Developers can change the default request behavior for all queries by setting `usingPostForQuery = true` to not cache any of the queries. Setting `usingPostForQuery = true` can be useful when using the Swift SDK on the server-side as many developers will not want the server caching queries.

### TODOs before merging

- [x] Add tests
- [x] Add entry to changelog

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,10 +6,10 @@ coverage:
   status:
     patch:
       default:
-        target: auto
+        target: 79
     changes: false
     project:
       default:
-        target: 87
+        target: 89
 comment:
   require_changes: true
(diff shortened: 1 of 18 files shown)


cbaker6:
@aliasad106 you can try out this branch for query cache. It should work without any changes needed on your end as long as you didn't change the default cache policy of the SDK. I've tested it with some of the query code in Playgrounds and it seems to work.

I still need to add test cases to ensure this doesn't break in the future.

PR #392, opened 2022-08-19
docs: add Parse ObjC SDK migration guide

mtrezza:

### Issue Description

Related issue: https://github.com/parse-community/Parse-Swift/issues/393

### Approach

Adds an early version of a migration guide from the Parse ObjC SDK.

### TODOs before merging
n/a

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,0 +1,41 @@
+# Migration from Parse ObjC SDK <!-- omit in toc -->
+
+This document describes how to migrate from the [Parse ObjC SDK](https://github.com/parse-community/Parse-SDK-iOS-OSX) to the Parse Swift SDK.
+
+ℹ️ *This document is a work-in-progress. If you find information missing, please submit a pull request to help us updating this document for the benefit of others.*
+
+- [Status of the SDKs](#status-of-the-sdks)
+- [Migration Instructions](#migration-instructions)
+- [Behavioral Differences](#behavioral-differences)
+- [Known Issues](#known-issues)
+- [Feature Comparison](#feature-comparison)
+
+# Status of the SDKs
+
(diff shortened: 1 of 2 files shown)


mtrezza:
@parse-community/swift-sdk Anything that should be added to this early version?

MIGRATION.md
```diff
@@ -0,0 +1,41 @@
...
+
+- x
+- x
+- x
+
+# Known Issues
+
+The issues below are important to consider before migrating.
+
+- ⚠️ Partially updating an object sends the full object to the server; this can have a significant impact on data transfer costs depending on your use case and architecture. All other Parse SDKs including the Parse ObjC SDK only send the changed properties to the server. For details see [GitHub issue #242](https://github.com/parse-community/Parse-Swift/issues/242).
```
cbaker6, line 33: This isn't true (#315) and anyone who follows the directions in [Playgrounds](https://github.com/parse-community/Parse-Swift/tree/main/ParseSwift.playground/Pages) for using the SDK will never encounter this.

cbaker6 (changes requested):

MIGRATION.md
mtrezza: Which directions specifically?

cbaker6: Below is incorrect:

> ⚠️ Partially updating an object sends the full object to the server; this can have a significant impact on data transfer costs depending on your use case and architecture.

A number of different things "can have a significant impact on data transfer costs..." on ALL of the SDK's which is usually due to bad design or incorrect use of the respective SDK. So this type of warning shouldn't be placed in any documents related to the Swift SDK unless you referring to all SDKs.

mtrezza: This is a migration guide. We need to look at it from from the perspective of a developer who migrates from the ObjC SDK and may expect the same or similar behavior from the Swift SDK, especially if all other Parse SDKs behave a certain way.

If saving an object, then updating a property and saving it again sends the full object to the server, *unless additional steps are taken which are not required in other Parse SDKs*, then this is something the developer should be made aware of.

mtrezza: Resolving since no further comment

mtrezza:
Merging this to get the draft out and give others the opportunity to extend it.

Issue #401, opened 2022-09-03
Send only changed properties to cloud by default when updating object


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).

### Current Limitation

The Parse Swift SDK requires additional code overhead and developer considerations when updating objects. Specifically, to send only changed properties to the server, the SDK requires the developer to create an object clone from `ParseObject.mergeable` and override an internal method `ParseObject.merge `. The same behavior is achieved by other Parse SDKs without these efforts.

The disadvantages are:
- Developer needs to manually maintain custom properties in the overridden `ParseObject.merge` method. For example, if a new custom property is added to the object but not to the internal method, it creates data inconsistencies that may lead to bugs that are difficult to track down.
- Adds code complexity for what is supposed to be one of the simplest operations - updating an object.

### Feature / Enhancement Description

The Parse Swift SDK should send only the changed properties to the server by default. Without requiring the developer to override an internal method of `ParseObject` and without having to clone the object from `ParseObject.mergeable`.

### Example Use Case

The following examples compare how to update a saved object in the Parse ObjC SDK vs. the Parse Swift SDK.

```swift
// Parse ObjC SDK
PFObject *obj = [PFObject objectWithClassName:@"Example"];
obj[@"key"] = @"value1";
[obj save];
obj[@"key"] = @"value2";
[obj save];

// Parse Swift SDK
struct Example: ParseObject {
  var objectId: String?
  var createdAt: Date?
  var updatedAt: Date?
  var ACL: ParseACL?
  var originalData: Data? 
  var key: String?

  func merge(with object: Self) throws -> Self { 
    var updated = try mergeParse(with: object) 
    if updated.shouldRestoreKey(\.key, original: object) { 
      updated.key = object.key 
    }
    return updated
  }
}

let obj = Example()
obj.key = "value1"
obj.save()
var objMergable = obj.mergeable
objMergable.key = "value2"
objMergable.save()
```

### Alternatives / Workarounds

n/a

### References

- https://github.com/parse-community/Parse-Swift/issues/242

mtrezza:
[...]

PR #406, opened 2022-09-07
feat: add set() and auto merge updated properties with original

cbaker6:

### Issue Description

Developers need to call the `mergeable` instance property when they start to make updates to an `ParseObject` in order to use the Parse Servers "quirky patch". To have a local copy of the updated ParseObject merged with original, developers also need to implement the `merge()` method. If `merge()` is not implemented, the developer can call `fetch()` to get the merged updates.
 
Related issue: #401 

### Approach

Build onto #315 and add a `set` method that developers can call on their objects which automatically creates a `mergeable` instance that can be merged with the original object without the developer implementing `merge()`. An example is below:

```swift
/*
Assume you have a ParseObject named GameScore that has
properties named points and levels.
*/ 
var score = GameScore()
score.points = 10
let savedScore = try await score.save() // Create ParseObject on server
/* 
Now you want to update the saved object.
Multiple operations can be chained together.
*/
var updatedScore = try savedScore
      .set(\.points, to: 15)
      .set(\.levels, to: [1])
let savedUpdatedScore = try await updatedScore.save()
print(savedUpdatedScore) // Updated score
```

If a developer **chooses** to not implement the `merge()` and depends on the new default merging in this PR they are choosing to incur more computational overhead on the respective client after the update is already saved. This is due to the extra encoding/decoding needed to discover and complete the merge. The overhead will most likely increase the larger the `ParseObject`. If a developer wants to skip the overhead, they should implement `merge()` on each of their `ParseObjects`. The new `set` method is below:

```swift
/**
     Set the value of a specific `KeyPath` on a `ParseObject`.
     - parameter key: The `KeyPath` of the value to set.
     - parameter to: The value to set the `KeyPath` to.
     - returns: The updated `ParseObject`.
     - important: This method should be used when updating a `ParseObject` that has already been saved to
     a Parse Server. You can also use this method on a new `ParseObject`'s that has not been saved to a Parse Server
     as long as the `objectId` of the respective `ParseObject` is **nil**.
     - attention: If you are using the `set()` method, you do not need to implement `merge()`. Using `set()`
     may perform slower than implementing `merge()` after saving the updated `ParseObject` to a Parse Server.
     This is due to neccesary overhead required to determine what keys have been updated. If a developer finds decoding
     updated `ParseObjects`'s to be slow, implementing `merge()` may speed up the process.
     - warning: This method should always be used when making the very first update/mutation to your `ParseObject`.
     Any subsequent mutations can modify the `ParseObject` property directly or use the `set()` method.
 */
    func set<W>(_ keyPath: WritableKeyPath<Self, W?>, to value: W) -> Self where W: Equatable
``` 

### TODOs before merging

- [x] Add tests
- [x] Add entry to changelog


mtrezza:
>  If you have add properties to your `ParseObject`that are of type `Date` and need the precise time of
     the date, you should implement `merge()` or call `fetch()` on your object after updating.

What does that mean? Could you give an example what the effect on a date value is?

cbaker6:
> What does that mean? Could you give an example what the effect on a date value is?

This is has been updated in the docs, it's not an issue.

mtrezza:
Could you point to where in the docs that is referred to?

mtrezza:
Great!

Issue #426, opened 2022-10-19
New installation object created when migrating from ObjC SDK


mtrezza:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-Swift/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Swift SDK](https://github.com/parse-community/Parse-Swift/releases). 

### Issue Description

After migrating an app to the Parse Swift SDK, launching the app will create a new `_Installation` object with a new `installationId`. It will appear as if the app had been uninstalled and then reinstalled, even though it was only updated with the new Parse Swift SDK.

### Steps to reproduce
n/a

### Actual Outcome

A new installation object is created.

### Expected Outcome

Any existing installation object should be used as the current installation.

### Environment

Client
- Parse Swift SDK version: `4.14.2`

cbaker6:
Those developers seeing this issue should note that this is not true and is not an issue. If you really want to migrate from the Objective-C SDK to the Swift SDK, see the convo, sample app, and a developer who successfully migrated [here](https://community.parseplatform.org/t/how-to-transition-from-the-old-parse-framework/2807/19?u=cbaker6).

Since I’ve already solved this problem, I will not be spending any additional time on Objective-C SDK migration, so those looking for a better way will need to depend on someone else to provide a different solution.

mtrezza:
If this is not an issue anymore, can we remove it from the migration guide?

mtrezza:
I've searched through code for instructions on how to migrate the installation:

```swift
let objcInstallation = PFInstallation.current()!
let swiftInstallation = try await Installation.become(objcInstallation.objectId!)
```

It seems to expect that the Parse ObjC SDK is installed in parallel to the Parse Swift SDK, is that correct?

cbaker6:
> It seems to expect that the Parse ObjC SDK is installed in parallel to the Parse Swift SDK, is that correct?

The description of https://github.com/parse-community/Parse-Swift/pull/407#issue-1368984914 states:

> Adding a method like this allows developers to use their own ways to get a particular objectId and use it to migrate to the Swift SDK.

As for:
[...]

PR #427, opened 2022-10-20
docs: add installation migration info

mtrezza:

### Issue Description

No instructions for installation migration.

Related issue: https://github.com/parse-community/Parse-Swift/issues/426

### Approach

Add instructions for installation migration.

### TODOs before merging
n/a

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -8,6 +8,7 @@ This document describes how to migrate from the [Parse ObjC SDK](https://github.
 - [Migration Instructions](#migration-instructions)
 - [Behavioral Differences](#behavioral-differences)
 - [Known Issues](#known-issues)
+  - [Installation](#installation)
 - [Feature Comparison](#feature-comparison)
 
 # Status of the SDKs
@@ -116,9 +117,26 @@ The Parse ObjC SDK will be phased out in the future in favor of the more modern 
 
 # Known Issues
 
-These issues below are especially important to consider before migrating.
+These following issues are important to consider before migrating.
 
-- After migrating an app to the Parse Swift SDK, launching the app will create a new `Installation` object with a new `installationId`. It will appear as if the app had been uninstalled and then reinstalled, even though it was only updated with the new Parse Swift SDK.
+## Installation
+
+After migrating an app to the Parse Swift SDK, launching the app will create a new `_Installation` object with a new `installationId`. It will appear as if the app had been uninstalled and then reinstalled, even though it was only updated with the new Parse Swift SDK.
+
+This may be problematic if the installation object is directly referenced in your app or if it contains fields that should be maintained. To address this, there are currently two options:
+
+  - a) Use the Parse Swift SDK *together with* the Parse ObjC SDK in your project. This way you can migrate the installation by referencing the installation's `objectId`:
+
+    ```swift
+    import Parse      // Parse ObjC SDK
+    import ParseSwift // Parse Swift SDK
+
+    let objcInstallation = PFInstallation.current()!
+    let swiftInstallation = try await Installation.become(objcInstallation.objectId!)
+    ```
+  - b) Release an update for your app which stores the installation's `objectId` somewhere. Once the app has seen a sufficient adoption, you could release an app update with only the Parse Swift SDK and use the stored `objectId` to migrate to the installation.
+
+Both options may require a long-term migration if your existing app has many installations and until the adoption rate with either approach is high enough.
 
 # Feature Comparison
 


MIGRATION.md
```diff
@@ -116,9 +117,26 @@ The Parse ObjC SDK will be phased out in the future in favor of the more modern
 
 # Known Issues
 
-These issues below are especially important to consider before migrating.
+The following issues are important to consider before migrating.
 
-- After migrating an app to the Parse Swift SDK, launching the app will create a new `Installation` object with a new `installationId`. It will appear as if the app had been uninstalled and then reinstalled, even though it was only updated with the new Parse Swift SDK.
+## Installation
+
+After migrating an app to the Parse Swift SDK, launching the app will create a new `_Installation` object with a new `installationId`. It will appear as if the app had been uninstalled and then reinstalled, even though it was only updated with the new Parse Swift SDK.
```
cbaker6, line 124: I believe there should be a clear distinction between `converting` and `migrating`. Converting an app from using the Objective-C SDK to Swift SDK shouldn’t assume migration. A migration typically entails some formal process or method(s) that are intentionally called to complete migration. If a developer wants to migrate a `PFUser` from the Objective-C Keychain to the Swift SDK Keychain, the developer needs to intentionally call [ParseUser.loginUsingObjCKeychain](https://parseplatform.org/Parse-Swift/release/documentation/parseswift/parseuser/loginusingobjckeychain(options:)/). Similarly, if the developer wants to migrate a `PFInstallation` from the Objective-C SDK, they need to intentionally call [ParseInstallation.become](https://parseplatform.org/Parse-Swift/release/documentation/parseswift/parseinstallation/become(_:copyentireinstallation:options:)). This Swift SDK was never intended to be a drop-in replacement for the Objective-C SDK, yet the way this migration document is structured tends to imply this is the case, it is not.

I don’t recall ever seeing comments or documentation about other Client SDK’s (Objective-C, JS, and Flutter) that work for IOS and other Apple OS’s state they can automatically access a ParseUser or ParseInstallation from another client SDK or imply they can be used interchangeably. If the other SDK’s aren’t held to this type of standard and don’t have comments or docs in their repos stating this, then why should the Swift SDK? In addition, these types of things should not be labeled as `issues` or `bugs` in this SDK.

> launching the app will create a new `_Installation` object with a new `installationId`.

In most of the client SDK’s, I don’t see `_Installation` used. `_Installation` is typically used on the server and in the dashboard. This comment is misleading as it makes it seem like the Swift SDK creates a new installation on the server. It does not do this. The installation is only local on the devices keychain. 

> It will appear as if the app had been uninstalled and then reinstalled, even though it was only updated with the new Parse Swift SDK.

This comment implies `drop-in replacement` which I addressed earlier. The Swift SDK isn’t a drop-in replacement and verbiage that eludes to such should be removed and/or modified as it’s misleading.

cbaker6 (changes requested):

MIGRATION.md
```diff
@@ -116,9 +117,26 @@ The Parse ObjC SDK will be phased out in the future in favor of the more modern
...
+
+This may be problematic if the installation object is directly referenced in your app or if it contains fields that should be maintained, like the `badge` number field for example. To address this, here are two options:
```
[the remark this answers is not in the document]
mtrezza: If the app references the current installation object for any purpose, that reference will be incorrect, because that's not the current installation anymore, even though the user didn't reinstall.

Same issue with data like the badge counter, which will suddenly become zero for example, which has an effect on the user as well.

MIGRATION.md
mtrezza: It covers the same iOS ecosystem, so people will migrate. That involves a process, which is described here.

mtrezza: It doesn't change much whether we call this a migration guide or a conversion guide - the issues stay the same. Let's stay with "migration" as that is maybe a more common term and it's already used to reference this document.

mtrezza: It doesn't change much whether we call this a migration guide or a conversion guide - the issues stay the same. Let's stay with "migration" as that is maybe a more common term and it's already used to reference this document.

The standpoint you represent here goes beyond the terminology. I invite you to have a separate chat about this, because I see some fundamental discrepancies between your standpoint and developer/business expectations.

It's important that we get this migration info out to the developers, so I will consider this resolved. If you have any specific rewording suggestions please suggest.

cbaker6:
You can handle adding and reviewing any future changes to this document since you tend to ignore any changes you don’t like or commit to the main branch whenever you want. In the other repos, you force others to conform to your reviews and changes and hold up the PR’s, but don’t hold yourself to the same standard. Others have complained about this as well. I have no other comments on this document or discussion related to migration, so don’t reference me when questions come up as it’s a waste of my time.

mtrezza:
If you have any specific suggestions I'm open to discuss, but you haven't put anything forward. Instead you are questioning wether we should call it "migration" or "conversion". It's beside the point of providing the community a central document for switching form ObjC to Swift SDK.

Developers are switching over from the ObjC SDK. The Swift SDK needs to address that, its not a walled-off standalone product. It's part of Parse Platform and it's our obligation to be mindful about the needs of developers.

mtrezza:
If you have any specific suggestions I'm open to discuss, but you haven't put anything forward. Instead you are opening fundamental discussions to which standard we should hold this SDK. The reality is that developers are switching over from the ObjC SDK. The Swift SDK needs to address that, its not a walled-off standalone product. It's part of Parse Platform and it's our obligation to be mindful about the needs of developers and provide a central document for switching from ObjC to Swift SDK.

PR #462, opened 2026-05-11
fix: Reuse Objective-C SDK installation on migration

### LINKED ISSUE #426: New installation object created when migrating from ObjC SDK

Branchrise:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-Swift/issues/426).

### Issue Description

When migrating from the Objective-C SDK, `migratingFromObjcSDK` copies the Objective-C Keychain `installationId` into the Swift SDK current installation, but it does not reuse the existing `_Installation` object from Parse Server. The current installation can therefore remain without the existing `objectId`, so a later save may create a new installation instead of using the migrated one.

Closes: https://github.com/parse-community/Parse-Swift/issues/426

### Approach

When the Objective-C Keychain contains an `installationId`, initialization now queries `_Installation` by that `installationId` using a reload cache policy. If an existing installation is found, it becomes the current installation and is persisted back to the Swift SDK storage/keychain, preserving the existing `objectId`.

If the lookup misses, the fallback still migrates the Objective-C `installationId`, but it avoids carrying over an unrelated saved `objectId` from the current Swift installation.

Added regression coverage for both the empty lookup fallback and the existing-installation reuse path.

### TODOs before merging

- none

### Testing

- `git diff --check`
- Attempted `swift test --filter InitializeSDKTests` locally with Swift 6.3.1 on Windows after installing the Swift toolchain and loading the Visual Studio Build Tools environment. The package currently fails before the targeted tests due to an existing Swift 6 parser error in `Sources/ParseSwift/Coding/ParseEncoder.swift`, outside this PR's changed files: `'let' binding pattern cannot appear in an expression`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Objective-C SDK migration: better detection of when to migrate, robust lookup and fallback behavior to ensure installation identifiers and existing installation data are synchronized and preserved.

* **Tests**
  * Expanded migration tests to cover lookup success, lookup failure, and preservation of pre-existing installations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/parse-community/Parse-Swift/pull/462)

### DIFF
--- a/Sources/ParseSwift/Parse.swift
+++ b/Sources/ParseSwift/Parse.swift
@@ -166,13 +166,14 @@ public func initialize(configuration: ParseConfiguration) {
 
         if shouldMigrateInstallation {
             var updatedInstallation = BaseParseInstallation.current
-            if var foundInstallation = try? BaseParseInstallation
-                .query("installationId" == installationId)
-                .first(options: [.cachePolicy(.reloadIgnoringLocalCacheData)]) {
+            do {
+                var foundInstallation = try BaseParseInstallation
+                    .query("installationId" == installationId)
+                    .first(options: [.cachePolicy(.reloadIgnoringLocalCacheData)])
                 foundInstallation.installationId = installationId
                 foundInstallation.updateAutomaticInfo()
                 updatedInstallation = foundInstallation
-            } else {
+            } catch let error as ParseError where error.code == .objectNotFound {
                 switch updatedInstallation {
                 case .some(let installation) where installation.objectId != nil:
                     updatedInstallation = BaseParseInstallation()
@@ -183,6 +184,9 @@ public func initialize(configuration: ParseConfiguration) {
                 }
                 updatedInstallation?.installationId = installationId
                 updatedInstallation?.updateAutomaticInfo()
+            } catch {
+                // initialize is non-throwing; keep the current installation untouched on lookup failures.
+                return
             }
 
             BaseParseInstallation.currentContainer.installationId = installationId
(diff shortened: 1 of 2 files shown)


Branchrise:
Addressed the latest CodeRabbit feedback in `55c133c1`.

- Replaced the Objective-C installation lookup `try?` with explicit `do`/`catch` handling.
- Only `ParseError.Code.objectNotFound` is treated as a lookup miss and allowed to fall back to a fresh installation.
- Other lookup failures leave the current installation untouched because `initialize` is non-throwing.
- Added a regression test for preserving the current installation when the lookup fails.

Validation:
- `git diff --check` passes.
- `swift test --filter InitializeSDKTests` is still blocked locally by the existing Swift 6 parser error in `Sources/ParseSwift/Coding/ParseEncoder.swift:839`, outside this PR's changed files.

PR #464, opened 2026-05-11
feat: Allow direct ParseObject partial updates

### LINKED ISSUE #408: Simplify updating object
### New Feature / Enhancement Checklist

- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-Swift/issues?q=is%3Aissue).

### Current Limitation

While https://github.com/parse-community/Parse-Swift/issues/401 with its PR https://github.com/parse-community/Parse-Swift/pull/406 was a great simplification when updating an object, it still requires special consideration when updating an object.

Specifically:
- a developer still has to keep in mind that there is a special approach with `obj.key.set(\.key, to: "value2")` (see [example](https://github.com/parse-community/Parse-Swift/issues/401#issuecomment-1242544040)) needed to only send updated keys to the server, and they cannot simply use `obj.key = "value2"`
- the special method contains a duplication of parameters, as `obj.key.set(\.key, to: "value2")` requires a reference to `key` two times

### Feature / Enhancement Description

Further simplify to allow to use `obj.key = "value2"` without calling any special method.

### Example Use Case

[Example](https://github.com/parse-community/Parse-Swift/issues/401#issuecomment-1242544040)

### Alternatives / Workarounds

Status quo

dclawd88:

- [x] I am not disclosing a vulnerability.
- [x] I am creating this PR in reference to an issue.

### Issue Description

Closes #408.

Saved or fetched `ParseObject` values currently need `mergeable` or `set()` to send only changed keys. Directly mutating a property on a saved value can still encode the unchanged fields into the update body because the SDK does not keep a per-object baseline for normal saved/fetched objects.

### Approach

- Store an `originalData` snapshot on objects returned from create, fetch, update, replace, and query responses.
- Keep `mergeable` compatible by reusing an existing snapshot when one is already present.
- Compare command-encoded object bodies against the stored snapshot and remove unchanged top-level keys before sending the request.
- Add coverage for direct property mutation and update existing expectations that saved updated objects retain a fresh snapshot.

### Validation

- `swift build` passes locally.
- `swift test --filter ParseObjectTests/testSaveUpdateCommandDirectPropertyMutation` could not run in this environment because only Command Line Tools are installed and the test target cannot import `XCTest` (`no such module 'XCTest'`).
- `xcodebuild` is unavailable because the active developer directory is `/Library/Developer/CommandLineTools` rather than a full Xcode install.

## Release Notes

* **New Features**
  * Optimized save requests to transmit only modified properties to the server, reducing network overhead
  * Added explicit support for delete operations on cleared properties in save requests

* **Bug Fixes**
  * Enhanced state preservation for objects retrieved from queries and save operations through improved snapshot tracking

[Review Change Stack](https://app.coderabbit.ai/change-stack/parse-community/Parse-Swift/pull/464)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,6 @@
 [Full Changelog](https://github.com/parse-community/Parse-Swift/compare/4.14.2...main), [Documentation](https://swiftpackageindex.com/parse-community/Parse-Swift/main/documentation/parseswift)
 * _Contributing to this repo? Add info about your change here to be included in the next release_
 
-__New features__
-- Allow saved ParseObjects to send only directly mutated properties without requiring `mergeable` or `set()`.
-
 ### 4.14.2
 [Full Changelog](https://github.com/parse-community/Parse-Swift/compare/4.14.1...4.14.2), [Documentation](https://swiftpackageindex.com/parse-community/Parse-Swift/4.14.2/documentation/parseswift)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 [Full Changelog](https://github.com/parse-community/Parse-Swift/compare/4.14.2...main), [Documentation](https://swiftpackageindex.com/parse-community/Parse-Swift/main/documentation/parseswift)
 * _Contributing to this repo? Add info about your change here to be included in the next release_
 
+__New features__
+- Allow saved ParseObjects to send only directly mutated properties without requiring `mergeable` or `set()`.
+
 ### 4.14.2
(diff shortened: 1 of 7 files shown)


dclawd88:
Ready for review. I addressed the CodeRabbit feedback in a follow-up commit: removed the manual changelog entry, avoided Swift 5.7-only `any ParseObject` syntax for the Swift 5.5 package, and added direct nil mutation handling via Parse Delete operations with coverage for the encoded update body. Local validation: `swift build` and `git diff --check` pass; `swift test` is blocked on this machine because the installed Command Line Tools environment cannot import XCTest.

